PR #5, opened 2018-03-03
some minor cleanup

jjmaldonis:
* cleaned up string formatting

* added language enums to use rather than strings

* better doc strings for classes to explain what they are used for

* moved some class definitions around in the storage file to better illustrate the hierarchy of the classes

* minor refactoring

Github makes it look like there are more changes than there really are. Big blocks of the storage.py file didn't get marked up correctly because I moved some of the classes around.


cdragontoolbox/data.py
```diff
@@ -0,0 +1,62 @@
+from enum import Enum
+
+
+class Region(Enum):
```
benoitryder: I don't see the purpose of an enum here.
Enum values and identifiers carry the same information, so it's just redundant. Having an enum rather than a plain string just make it more complicated to manipulate.

All we need is a list of known regions/langagues. Something like this:
```python
regions = ['br', 'cn', 'eun', ...]
```
It provides the basic features that we need: `some_string in regions`, `for region in regions`, `len(regions)`.

cdragontoolbox/export.py
```diff
@@ -124,7 +125,7 @@ def __init__(self, storage: Storage, output: str, stored=True):
                 if not versions:
                     break
         else:
-            raise ValueError("versions not found: %r" % versions)
+            raise ValueError(f'versions not found: {repr(versions)}')
```
benoitryder: You can use `{value!r}` in format string to format `value` using `repr()`.
(Multiple occurences.)

cdragontoolbox/export.py
```diff
@@ -175,9 +176,9 @@ def export(self):
         """
 
         if self.previous_patch:
-            logger.info("exporting patch %s based on patch %s", self.patch.version, self.previous_patch.version)
```
benoitryder: There is a reason for using `logger.info(fmt, arg1, arg2)` instead of a format string.
The message will only be formatted if the log is actually handled and output. For instance, debug messages will not be formatted if debug messages are dropped (which is the default).
Also, the logging API allows to define custom handlers, LogRecord classes, etc. that could make use of the args being separated.

Even if in our case the only would be a (negligeable) performance impact, I think it's good habit to keep the args separated.
What's your opinion?

cdragontoolbox/storage.py
```diff
@@ -163,187 +119,46 @@ def stream(self, urlpath) -> RequestStreamReader:
...
+        for name in os.listdir(base):
+            if os.path.isdir(f'{base}/{name}/releases'):
+                ret.append(Solution(self, name))
+        return ret
 
 
 class Solution:
     """
-    RADS solution
+    The Riot Application Distribution System (RADS) has two Solutions: `league_client_sln` and `lol_game_client_sln`.
```
benoitryder: Technically, more solutions would be supported and the code doesn't depend on it.
I would prefer to document the concrete use case in the README (I already had some words about solutions and projects), or make it clear that the code doesn't depend on it.
Actually... there are some pieces of code that depend on the actual solution names for parsing patch versions and exporting files, but not in Solution/Project.

(Same comment for Project.)

cdragontoolbox/wad.py
```diff
@@ -120,7 +121,13 @@ def guess_extension(data):
 
 class Wad:
     """
-    WAD archive
+    A WAD archive is a file that contains other files. It has a header that describes the format. There are multiple
```
benoitryder: Multi-line docstrings should start with a single-line summary of the documented object.
See: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

cdragontoolbox/storage.py
jjmaldonis: It looks like `SolutionVersion.patch_version` won't work for other solutions, but everything else should. I added a line in the doc string mentioning that: "These classes will likely work with other solutions, although some functionality may need to be extended."

I understand why it's useful to write these classes for general use cases, but their methods are pretty specific in some cases. I think it's nice to make it clear to the user what solutions we know we handle correctly, and therefore what their options are (unless they are doing something weird that we don't intend the classes to be used for).

cdragontoolbox/__init__.py
```diff
@@ -1,2 +1,20 @@
 import logging
 logger = logging.getLogger(__name__)
+
+from cdragontoolbox.storage import (
+    Version,
+    Storage,
+    Project, ProjectVersion,
+    Solution, SolutionVersion,
+    PatchVersion,
+    parse_component,
```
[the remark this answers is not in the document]
jjmaldonis: Fixed

cdragontoolbox/data.py
jjmaldonis: Fixed

cdragontoolbox/export.py
jjmaldonis: Fixed

cdragontoolbox/wad.py
jjmaldonis: Fixed

cdragontoolbox/export.py
jjmaldonis: This is kinda a tough one, but I think that the formatting of a string is such an insignificant time contribution to this code in particular that we shouldn't worry about it. f-strings are just so nice -- much much nicer than the `%` format, and it looks weird to be mixing old and new formatting options (I changed some of them then decided on this).

It would be different if the arguments were not evaluated unless the logger was used, but they have to be evaluated to read the "outer" method scope.

If people are using complex logging via custom handlers etc, then imo they can either submit a PR with any changes they need to just fork the repo and make their changes there. I think it's very unlikely that anyone is going to be using this code as-is in an app that requires customizable logging.

benoitryder: Resolved

tests/test_storage.py
```diff
@@ -1,6 +1,6 @@
 import os
 import pytest
-from tools import *
+from .tools import *
```
benoitryder: The `test` directory is not a package. You cannot use a relative import here.

PR #12, opened 2018-12-26
(WIP) Update/cleanup guessing methods 

Morilli:
A work in progress suggestion on some additions and changes to the current functions. Will likely add and modify a lot of stuff in this PR over time; this is just the first general preview of what I have currently. Feedback and critic is very welcome.

### DIFF
--- a/cdragontoolbox/__main__.py
+++ b/cdragontoolbox/__main__.py
@@ -184,13 +184,14 @@ def command_hashes_guess(parser, args):
     all_methods = [
         ("grep", "search for hashes in WAD files"),
         ("numbers", "substitute numbers in basenames"),
+        ("basenames", "substitute basenames"),
         ("words", "substitute known words in basenames"),
         ("ext", "substitute extensions"),
         ("regionlang", "substitute region and lang (LCU only)"),
         ("plugin", "substitute plugin name (LCU only)"),
         ("skin-num", "substitute skinNN numbers (game only)"),
         ("character", "substitute character name (game only)"),
-        ("prefixes", "substitute prefixes (game only)"),
+        ("prefixes", "check basename prefixes (game only)"),
     ]
     all_method_names = [name for name, _ in all_methods]
 
@@ -201,7 +202,7 @@ def command_hashes_guess(parser, args):
         return
 
     if not args.methods:
-        method_names = all_method_names
+        method_names = [name for name in all_method_names if name != "basenames" and name != "words"]
     else:
         method_names = [s.strip() for s in args.methods.split(',')]
         for name in method_names:
@@ -228,6 +229,8 @@ def command_hashes_guess(parser, args):
                 guesser.grep_wad(wad)
         if "numbers" in method_names:
             guesser.substitute_numbers()
+        if "basenames" in method_names:
+            guesser.substitute_basenames()
         if "words" in method_names:
             guesser.substitute_basename_words()
         if "ext" in method_names:
@@ -253,6 +256,8 @@ def command_hashes_guess(parser, args):
                 guesser.grep_wad(wad)
         if "numbers" in method_names:
             guesser.substitute_numbers()
+        if "basenames" in method_names:
+            guesser.substitute_basenames()
         if "words" in method_names:
             guesser.substitute_basename_words()
         if "ext" in method_names:
@@ -262,7 +267,7 @@ def command_hashes_guess(parser, args):
         if "character" in method_names:
             guesser.substitute_character()
         if "prefixes" in method_names:
-            guesser.substitute_basename_prefixes()
+            guesser.check_basename_prefixes()
 
         nfound = nunknown - len(guesser.unknown)
         if nfound:
@@ -442,7 +447,7 @@ def create_parser():
     subparser.add_argument('-n', '--dry-run', action='store_true',
                            help="list new hashes but don't update the hashes file")
     subparser.add_argument('-m', '--methods',
-                           help="list of guessing methods to run, comma-separated (default: all)")
+                           help="list of guessing methods to run, comma-separated (default: all except \"basenames\" and \"words\")")
     subparser.add_argument('--list-methods', action='store_true',
                            help="display a list of valid guessing methods and exit")
     subparser.add_argument('wad', nargs='*',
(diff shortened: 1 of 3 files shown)


cdragontoolbox/hashes.py
```diff
@@ -667,6 +700,19 @@ def grep_wad(self, wad):
 
         with open(wad.path, 'rb') as f:
             for wadfile in wad.files:
+                if wadfile.path_hash == 2155072684501898278 or wadfile.path == 10561690728639675755: # quality coding
```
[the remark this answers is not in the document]
Morilli: This is currently in place as these two files contain all included files of the DATA.wad.client and the Scripts.wad.client file. These have 4 byte prefixes and are currently not checked at all by grep_wad. I should probably figure out a way of making this a more general check, as I feel like this is a kinda sloppy check.

cdragontoolbox/hashes.py
```diff
@@ -289,13 +309,7 @@ def from_wads(cls, wads):
...
 
     def build_wordlist(self):
-        re_filter_path = re.compile(r"""'(?:
-            ^(?:plugins/rcp-be-lol-game-data/global/default/v1/champion-ability-icons/
-              | plugins/rcp-be-lol-game-data/global/default/data/characters/
-              )
-            | /[0-9a-f]{32}\.
-            )""", re.VERBOSE)
-        paths = (p for p in self.known.values() if not re_filter_path.search(p))
+        paths = (p for p in self.known.values())
```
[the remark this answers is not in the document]
Morilli: I actually misunderstood what this was trying to do, so I will re-add the ``data/characters`` part as well, as it reduces the wordlist by quite a bit. This should now also work again (didn't work before cause of a wrongly placed ``'``.

Morilli:
I think I should have cleared most stuff up. Please tell me if there is still something wrong with my changes. Also keep in mind that this is still a WIP, I will have to look at the grep_wad function to see what can be done there.

cdragontoolbox/hashes.py
```diff
@@ -667,6 +695,19 @@ def grep_wad(self, wad):
...
+                        path = m.group(0).lower().decode('ascii')
+                        pos = m.start()
+                        n = struct.unpack('<L', data[pos-4:pos])[0]
+                        paths.add(path[:n])
+                    for p in paths:
+                        if p.endswith('.lua'):
+                            self.check(p[:-4] + '.luabin')
+                        else:
+                            self.check(p)
                 if wadfile.ext in ('bin', 'inibin'):
```
benoitryder: Should become an `elif`.

cdragontoolbox/hashes.py
```diff
@@ -495,15 +508,31 @@ def substitute_numbers(self, nmax=100, digits=None):
         paths = self.known.values()
         super()._substitute_numbers(paths, nmax, digits)
 
+    def check_basename_prefixes(self, prefixes=None):
+        """Checks a provided list of prefixes for all basenames.
+        If no list is provided, a default one will be used"""
+        values = set()
+        if prefixes == None:
```
benoitryder: Use `is None` instead of `== None`.

cdragontoolbox/hashes.py
```diff
@@ -495,15 +508,31 @@ def substitute_numbers(self, nmax=100, digits=None):
...
+            prefixes = {"2x_","2x_sd_","4x_","4x_sd_","sd_"}
```
benoitryder: Add spaces after commas.

cdragontoolbox/hashes.py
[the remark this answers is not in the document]
benoitryder: Until those hashes are resolved I see two possibilities:
- keep hashes hardcoded as you suggested
- use `grep_file()` as fallback for unknown/generic binary files

The second solution would require to filter formats known to not contain paths (such as .dds). Code would be something like this:
```py
if wadfile.ext in ('bin', 'inibin'):
    # bin-specific processing
elif wadfile.ext == 'preload':
    # preload-specific processing
...
elif wadfile.ext in ('dds', 'tga', ...):
    pass  # don't grep filetypes known to not contain paths
else:
    # fallback: search for path-looking strings
    self.grep_file(data=wadfile.read_data(f))
```

Then update `grep_file()` to also support data rather than only a path (better than copy-pasting the implementation):
```py
def grep_file(self, path=None, data=None):
    if path is None:
        with open(path, "rb") as f:
            data = f.read()
    elif data is None:
        raise TypeError("either path or data must be provided")
    ....
```

Morilli:
I have applied the latest suggestions; the PR should be good to go now from my side.

cdragontoolbox/hashes.py
```diff
@@ -495,15 +508,31 @@ def substitute_numbers(self, nmax=100, digits=None):
         paths = self.known.values()
         super()._substitute_numbers(paths, nmax, digits)
 
+    def check_basename_prefixes(self, prefixes=None):
+        """Checks a provided list of prefixes for all basenames.
+        If no list is provided, a default one will be used"""
+        values = set()
+        if prefixes is None:
+            prefixes = {'2x_', '2x_sd_', '4x_', '4x_sd_', 'sd_'}
```
stelar7: Could this be passed as the default param value instead?

Morilli: You mean like `` def check_basename_prefixes(self, prefixes={'2x_', '2x_sd_', '4x_', '4x_sd_', 'sd_'}):``?

benoitryder: It's usually better to avoid using mutable objects as default parameters, because the same reference will be used each time.
For instance:
```py
def f(x=[]):
    x.append(1)
    return x
a = f()
b = f()
assert a is b
assert a == b == [1, 1]
```

In our case, `prefixes` is not modified by the function, so it would work. But it's easy to forget when updating the code and bugs could be introduced later on.

So IMHO Morilli's approach is the good one.

cdragontoolbox/__main__.py
```diff
@@ -201,7 +202,7 @@ def command_hashes_guess(parser, args):
         return
 
     if not args.methods:
-        method_names = all_method_names
+        method_names = [name for name in all_method_names if name != "basenames" and name != "words"]
```
benoitryder: Youd could use `if name not in ("basenames", "words")`.

PR #13, opened 2019-01-22
Add 2 new methods to hashes.py and other minor changes

Morilli:
Again, this is probably not perfect, but the general idea should be pretty clear. Also has support for the new ``.luabin64``™ files.


cdragontoolbox/hashes.py
```diff
@@ -328,6 +328,39 @@ def substitute_basename_words(self):
     def add_basename_word(self):
         super()._add_basename_word(self.build_wordlist())
 
+    def multi_substitution(self, plugin, fileext=""):
+        """Replaces a word in all basenames while adding an additional word near it.
+        Because of high running time, a specific plugin must and a file extension can be specified"""
```
benoitryder: Words are missing after `must`

cdragontoolbox/hashes.py
```diff
@@ -328,6 +328,39 @@ def substitute_basename_words(self):
     def add_basename_word(self):
         super()._add_basename_word(self.build_wordlist())
 
+    def multi_substitution(self, plugin, fileext=""):
```
benoitryder: Instead of filtering on a plugin name you should filter on a custom filter, or accept a custom list of paths as parameter. Otherwise, the method is LCU-specific.

cdragontoolbox/hashes.py
```diff
@@ -328,6 +328,39 @@ def substitute_basename_words(self):
...
+        """Replaces a word in all basenames while adding an additional word near it.
```
benoitryder: As said previously on Discord, this method should be made more generic. Current behavior seems very specific.

cdragontoolbox/hashes.py
```diff
@@ -382,8 +415,8 @@ def grep_wad(self, wad):
                     # splash config
                     # try to guess subdirectory name (names should only contain one element)
                     names = {s for path in jdata['files'].values() for s in re.findall(r'-splash-([^.]+)', path)}
```
benoitryder: Factorize the `lower()` call here (use `s.lower()`).

cdragontoolbox/hashes.py
```diff
@@ -406,8 +439,8 @@ def grep_wad(self, wad):
 
             # relative path starting with ./ or ../ (e.g. require() use)
             relpaths |= {m.group(1) for m in re.finditer(r'[^a-zA-Z0-9/_.\\-]((?:\.|\.\.)/[a-zA-Z0-9/_.-]+)', data)}
-            # basename or subpath (check for a extension)
-            relpaths |= {m.group(1) for m in re.finditer(r'''["']([a-zA-Z0-9][a-zA-Z0-9/_.@-]*\.(?:js|json|webm|html|[a-z]{3}))\b''', data)}
+            # basename or subpath (check for an extension)
+            relpaths |= {m.group(1) for m in re.finditer(r'''["']([a-zA-Z0-9][a-zA-Z0-9/_.@-]*\.(?:js|json|webm|html|[a-z]{3}))["']''', data)}
```
benoitryder: `\b` is more generic and handles more cases.

cdragontoolbox/hashes.py
Morilli: The method is indeed currently lcu specific, that's why it's only in the lcu hash guesser.

cdragontoolbox/hashes.py
Morilli: This is meant as in "a specific plugin must be specified and a file extension can be specified", just in shorter.

cdragontoolbox/hashes.py
benoitryder: If you make it generic, you could move it to the be class.

cdragontoolbox/hashes.py
Morilli: Do you have an example for a file where neither ``"`` nor ``'`` is used for a relative path? Cause I could not imagine such a case and I feel like my change makes it logically more intuitive.

benoitryder: I don't remember exactly where but I think I had such cases. You could imagine any separator at the end of the path.

Morilli:
I made some proof of concept changes to the multi_substitution function. It should preserve the current functionality for lcu but make providing a custom list of paths and words possibly. Please review on whether this is what you asked for or whether I should be using a different approach altogether.

cdragontoolbox/hashes.py
```diff
@@ -328,6 +341,35 @@ def substitute_basename_words(self):
...
         super()._add_basename_word(self.build_wordlist())
 
+    def multi_substitution(self, plugin, fileext=""):
+        """Builds a list of paths based on the specified plugin and file-extension.
+        Replaces a word in all basenames in these paths while adding an additional word near it."""
+
+        words = self.build_wordlist()
+        paths = set()
+        for path in self.known.values():
+            if path.startswith(f"plugins/{plugin}") and path.endswith(fileext):
```
benoitryder: - You can factorize `f"plugins/{plugin}"`.
- `paths` can be a list (there will be no duplicates to remove)

benoitryder:
Customization of paths and words is good.
What about the customization of "changes" (number of words to add for instance)?

benoitryder:
Yep! all good!

cdragontoolbox/hashes.py
```diff
@@ -185,8 +185,13 @@ def check_xdbg_hashes(self, path):
     def check_basenames(self, names):
         """Check a list of basenames for each known subdirectory"""
 
-        dirs = self.directory_list()
-        self.check_iter(f"{d}/{name}" for name in names for d in dirs)
+        dirs = list(self.directory_list())
+        unknown = self.unknown # global -> local for increased performance
+        for name in progress_iterator(sorted(names)):
+            for dir in dirs:
+                h = xxh64(name).intdigest() # inline check() for increased performance
```
benoitryder: Optimizations should not affect the readibility and maintainability of this code.
Pure Python code is not meant to, and never will, be incredibly fast. If you aim for speed, you should use another approach/language.

cdragontoolbox/hashes.py
```diff
@@ -205,38 +210,83 @@ def substitute_basenames(self):
...
+                    self._add_known(h, name)
+
+    def _substitute_basename_words(self, paths, words, amount=1):
+        """Substitutes {amount} side by side words in all basenames
+        of all given paths (default=1 for simple 1 word substitution)."""
+
+        regex_part = ""
+        format_part = ""
+        for i in range(1, amount):
+            regex_part += r'([-_][^/_.-]+)'
```
benoitryder: - `regex_part = r'...' * (amount-1)` (same for `format_part`)
- `.` is a valid separator used in some basenames

cdragontoolbox/hashes.py
```diff
@@ -205,38 +210,83 @@ def substitute_basenames(self):
...
+            format_part += r'{sep}%%s'
+        format_part = ''.join([r'%s%%s', format_part, r'%s'])
```
benoitryder: Just use `format_part = f'%s%%s{format_part}%s'`

Same for other similar uses of `'.join()`.

cdragontoolbox/hashes.py
```diff
@@ -205,38 +210,83 @@ def substitute_basenames(self):
...
+        re_extract = re.compile(''.join([r'([^/_.-]+)(?=', regex_part, r'[^/]*\.[^/]+$)']))
+        temp_formats = set()
+        for path in paths:
+            for m in re_extract.finditer(path):
+                match = ""
+                for i in range(amount):
+                    match += m.group(i+1)
```
benoitryder: `match = ''.join(m.group(i+1 for i in range(amount))`

cdragontoolbox/hashes.py
```diff
@@ -205,38 +210,83 @@ def substitute_basenames(self):
...
+        """Substitutes {amount} side by side words in all basenames
+        of all given paths (default=1 for simple 1 word substitution)."""
+
+        regex_part = ""
+        format_part = ""
+        for i in range(1, amount):
+            regex_part += r'([-_][^/_.-]+)'
+            format_part += r'{sep}%%s'
+        format_part = ''.join([r'%s%%s', format_part, r'%s'])
+        re_extract = re.compile(''.join([r'([^/_.-]+)(?=', regex_part, r'[^/]*\.[^/]+$)']))
```
benoitryder: Why not simply using something like `((?:[-_][^/_.-]+){amount})`?

Here you only split the individual words just to join them just afterwards.

cdragontoolbox/hashes.py
```diff
@@ -205,38 +210,83 @@ def substitute_basenames(self):
...
 
-        re_extract = re.compile(r'([^/_.-]+)(?=[^/]*\.[^/]+$)')
         formats = set()
-        for path in self.known.values():
-            for m in re_extract.finditer(path):
-                formats.add('%s%%s%s' % (path[:m.start()], path[m.end():]))
+        for format in temp_formats:
+            formats.update(format.replace("{sep}", sep) for sep in "-_")
 
+        product = [prod for prod in itertools.product(words, repeat=amount)]
```
benoitryder: This create a possibly very large list of words. I don't think it's a good idea to factorize it, memory-wise.

cdragontoolbox/hashes.py
```diff
@@ -205,38 +210,83 @@ def substitute_basenames(self):
...
+
+    def _substitute_basename_words(self, paths, words, amount=1):
+        """Substitutes {amount} side by side words in all basenames
+        of all given paths (default=1 for simple 1 word substitution)."""
+
+        regex_part = ""
+        format_part = ""
+        for i in range(1, amount):
+            regex_part += r'([-_][^/_.-]+)'
+            format_part += r'{sep}%%s'
```
benoitryder: No need for a raw string here/

cdragontoolbox/hashes.py
```diff
@@ -205,38 +210,83 @@ def substitute_basenames(self):
...
+                    match += m.group(i+1)
+                temp_formats.add(format_part % (path[:m.start()], path[m.span()[0]+len(match):]))
 
-        re_extract = re.compile(r'([^/_.-]+)(?=[^/]*\.[^/]+$)')
         formats = set()
-        for path in self.known.values():
-            for m in re_extract.finditer(path):
-                formats.add('%s%%s%s' % (path[:m.start()], path[m.end():]))
+        for format in temp_formats:
+            formats.update(format.replace("{sep}", sep) for sep in "-_")
```
benoitryder: This oneline is not that longer: `formats = {fmt.replace("{sep}", sep) for fmt in temp_formats for sep in "-_")}

cdragontoolbox/hashes.py
```diff
@@ -315,18 +369,34 @@ def substitute_region_lang(self):
...
         region_lang_list = [(r, l) for r in regions for l in langs]
         known = list(self.known.values())
+        logger.info(f"substitute region and lang")
         for region_lang in progress_iterator(region_lang_list, lambda rl: f"{rl[0]}/{rl[1]}"):
             replacement = r'plugins/\1/%s/%s/' % region_lang
             self.check_iter(regex.sub(replacement, p) for p in known)
 
-    def substitute_basename_words(self):
-        super()._substitute_basename_words(self.build_wordlist())
+    def substitute_basename_words(self, plugin="", fileext="", words=None, amount=1):
```
benoitryder: Use `None` instead of `""` for *no value provided*.

cdragontoolbox/hashes.py
Morilli: That's right, didn't think about that.

cdragontoolbox/hashes.py
benoitryder: You can use something like that:
```py
paths = list(self.known.values())
if plugin is not None:
    paths = [path for path in paths if path.startswith(f"plugins/{plugin}/")]
if fileext is not None:
    paths = [path for path in paths if path.endswith(fileext)]
```

cdragontoolbox/hashes.py
Morilli: I'd argue cause the ``check()`` is only 3 lines and it is the only thing being executed, readability isn't really affected by writing it out like that. And considering that performance is increased by around 30% to before (if I'm not mistaken) I think having it like this is fine.

benoitryder: You should fix the checked path (`f"{dir}/{name}"` instead of just `name`) before comparing the performance. ;)

And yes, readibility IS affected because you have 5 lines that use another module and "internal" instance variables instead of just 1 call to a single function with a straightforward name.

Morilli: Checking basenames ``f"{a}.png" for a in wordlist`` results in 22 seconds with a simple ``self.check()``, while having the three lines (and the ``unknown = self.unknown``) in there manually results in 17 seconds. These results are pretty consistent and show that performance increase is indeed at around 30%. Factoring out dots is also a pretty fine thing to do imo, it's even done in ``check_iter()`` for increased performance.

cdragontoolbox/hashes.py
benoitryder: If you decide to exclude the `.suffix.ext` trailing part used by some files, then you could simplify the regexp since you shouldn't need the `[^/]*\.[^/]+$` part.

Also, since you modified the regexp, you need to update the use of `m.group()` to: all parts are in the same group so you don't need to join them anymore.

cdragontoolbox/hashes.py
benoitryder: So, you have a ~22% performance increase with `f"{a}.png`, which is a simpler use case (shorter string, single format variable). Even if you reach 20% performance increase it's still nothing compared to what you could achieve with a perfomance-oriented language or interpreter.

My point is: if you want performance, aim for it and do it the right way. Instead of manually inlining Python code to gain ~20%, use the proper tools and gain ~200% instead.

benoitryder: They are valid methods if your main objective is to improve performance of your Python code. But it's not the case of our code.

Improving performance often comes with a trade-off. And when you inline `check_iter()` and `check()` whenever they are used, the trade-off is not worth it.

benoitryder: The main goal is to have automatic guessing for new patches that will find a reasonnably good amount of new hashes in a reasonnable amount of time, while keeping the code clean, readable and maintainable. Here you barely achieve a 20% performance gain (once again, not 30̈% according to your numbers) for an operation that is already fast enough. Notably, the cost of `check_iter()` is nothing compared to the cost of extracting files from WADs to grep them.

Your goal is to bruteforce hashes as fast as possible, it's not the same goal. If you want/need additional guessing methods you can add them to your own script and implement them as you want.

And yes, maintainability and readibility are reduced because checking hashes is not factorized anymore. So if the implementation of `check()`, `check_iter()` or some `HashGuesser' internals change, you'll have to check every function and modify it if needed.

Morilli: You're right about the factorization part; the code will have to be manually adjusted every time the ``check()`` code changes. I find the probability for that to happen to be very low though, and even if we were to make minor adjustments; replacing 6 occurences in the code with these surely isn't gonna be an hourly long process.

The functions I am optimizing in this PR are not functions that are frequently run on new patch releases; I am optimizing functions that are specifically here to provide brute forcing functionality for hashes that can't be found by grepping files. Surely the automatic guessing functionality isn't the only focus, else there would not be other brute forcing functions in the first place.

Also I don't see how you're getting your 20% performance increase value; from recent tests I have a running time of 76 seconds for ``substitute_basenames()`` on the old version, and 59 seconds for my PR'd version. 76/59 ≈ 1.29, which means 29% faster, not 20%.

benoitryder: You said time went down from 22s to 17s. Those are the numbers you gave.

Morilli: And 22/17 is 1.294, which again is ~ 30%+.

benoitryder: You're right for the numbers, my bad. I wrongly swapped the two values.

Regarding the inlining, it's not about which method is frequently run or not, it's about code clarity, consistency. As I already said, if your goal is to provide functions *specifically here to provide brute forcing*, you are doing it wrong, because performance will be bad anyway compared to what you could **easily** achieve using a suitable language (e.g. C++).

Morilli: > if your goal is to provide functions specifically here to provide brute forcing, you are doing it wrong

We have numerous functions that rely on brute forcing. Of course having this entire thing be written in a language that supports more optimisations like C or C++ is generally a good idea, it's just currently not the case. For the time being I don't any problem regarding brute forcing functions, especially cause they aren't simple "test this string" functions but are rather manipulating already found strings to guess new ones. 

About the code clarity/consistency stuff:
Would you be fine with keeping the self.check_iter(...) in the functions like before but changing the check_iter function to look like this?
```python
    def check_iter(self, paths):
        """Check paths from an iterable"""

        # failsafe for common dumb error
        if isinstance(paths, str):
            raise TypeError("expected iterable of strings, got a string")
        unknown = self.unknown
        for p in paths:
            h = xxh64(p).intdigest()
            if h in unknown:
                self._add_known(h, p)
```

benoitryder: Yes. If you want to optimize `check_iter()`, it's fine.

Morilli:
I *think* I covered everything now, tell me if something is still wrong / unwanted.

cdragontoolbox/hashes.py
```diff
@@ -198,45 +201,69 @@ def directory_list(self, cached=True):
...
         re_extract = re.compile(r'([^/_.-]+)(?=[^/]*\.[^/]+$)')
         formats = set()
-        for path in self.known.values():
-            if "assets/characters" not in path and "vo/" not in path and "sfx/" not in path and "skins_skin" not in path: # Reduce the iteration time for game hashes
-                for m in re_extract.finditer(path):
-                    formats.update('%s%%s%s%s' % (path[:m.start()], sep, path[m.start():]) for sep in "-._")
-                    formats.update('%s%s%%s%s' % (path[:m.end()], sep, path[m.end():]) for sep in "-._")
+        for path in paths:
+            for m in re_extract.finditer(path):
+                formats.update('%s%%s%s%%s%s' % (path[:m.start()], sep, path[m.end():]) for sep in "-_")
```
benoitryder: You can easily make this method generic by replacing with any required number of `%%s` then use `product(words, repeat=N)`.

cdragontoolbox/hashes.py
```diff
@@ -381,17 +431,17 @@ def grep_wad(self, wad):
                 elif 'musicVolume' in jdata and 'files' in jdata:
                     # splash config
                     # try to guess subdirectory name (names should only contain one element)
-                    names = {s for path in jdata['files'].values() for s in re.findall(r'-splash-([^.]+)', path)}
+                    names = {s.lower() for path in jdata['files'].values() for s in re.findall(r'-splash-([^.]+)', path)}
                     self.check_iter(f"plugins/rcp-fe-lol-splash/global/default/splash-assets/{name}/config.json" for name in names)
-                    self.check_iter(f"plugins/rcp-fe-lol-splash/global/default/splash-assets/{name}/{path}" for name in names for path in jdata['files'].values())
+                    self.check_iter(f"plugins/rcp-fe-lol-splash/global/default/splash-assets/{lower}/{path}" for name in names for path in jdata['files'].values())
```
benoitryder: `lower` -> `name`

cdragontoolbox/hashes.py
```diff
@@ -744,19 +801,20 @@ def grep_file(self, path=None, data=None):
 
         # find path-like strings, then try to parse the length
         paths = set()
-        for m in re.finditer(br'(?:ASSETS|DATA|LEVELS)/[0-9a-zA-Z_. /-]+', data):
+        for m in re.finditer(br'(?:ASSETS|DATA|DATA_SOON|LEVELS)/[0-9a-zA-Z_. /-]+', data):
             path = m.group(0).lower().decode('ascii')
-            paths.add(path)
+            paths.add(path.replace("data_soon","data"))
```
benoitryder: Add a space after commas (same below).

cdragontoolbox/hashes.py
```diff
@@ -198,45 +201,69 @@ def directory_list(self, cached=True):
...
     def substitute_basenames(self):
         """Check all basenames in each subdirectory"""
 
-        names = set(os.path.basename(p) for p in self.known.values())
+        names = {os.path.basename(p) for p in self.known.values()}
         dirs = self.directory_list()
         logger.info(f"substitute basenames: {len(names)} basenames, {len(dirs)} directories")
         for name in progress_iterator(sorted(names)):
-            self.check_iter(f"{d}/{name}" for d in dirs)
+            self.check_iter(f"{dir}/{name}" for dir in dirs)
```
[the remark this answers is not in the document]
Morilli: This was left in from my previous restructuring. I feel like it looks a bit more logical now, but I can change it back if that's unwanted.

cdragontoolbox/hashes.py
benoitryder: The main use for y>2 is to enlarge the set of tested paths. If you search for specific hashes you start with y=1, then try with y=2 if the first run didn't yield any result, etc.

Morilli: I'm just not sure whether I should just make a single function that offers support for all combinations of x and y, and have name links from other functions; like ``substitute_basename_words`` calling that main function with the specified amount of substitutions.

benoitryder: A method that allows any combination of x and y might be a bit complicated.
If you feel confident with it, that would be great. But a `multiple_substitution()` method that offers `x = 1, y >= 2` is not hard to implement by extending `double_substitution()` and would be good enough IMO.

Morilli: It's not that difficult as ``substitute_basename_words`` already supported varying x and y, with x = y as condition. I've not split these two values. Please review the structure of the input variables.

cdragontoolbox/hashes.py
```diff
@@ -213,14 +213,17 @@ def substitute_basenames(self):
...
-    def _substitute_basename_words(self, paths, words, amount=1):
-        """Substitutes {amount} side by side words in all basenames
-        of all given paths (default=1 for simple 1 word substitution)."""
+    def _substitute_basename_words(self, paths, words, x=1, y=1):
+        """Replaces x side by side words with y words in all basenames of all given paths
+        (default=1 for simple 1 word substitution). x > 0; y > 0 is required."""
 
+        if not (x and y):
+            print("Either x or y is 0.")
+            return
```
benoitryder: If arguments are invalid, raise a ValueError instead of printing.

cdragontoolbox/hashes.py
```diff
@@ -349,36 +336,26 @@ def substitute_region_lang(self):
...
-    def substitute_basename_words(self, plugin=None, fileext=None, words=None, amount=1):
-        """Replaces {amount} side by side words in all basenames (default=1 for simple 1 word substitution).
+    def substitute_basename_words(self, plugin=None, fileext=None, words=None, amount=None, x=1, y=1):
+        """Substitutes {amount} side by side words in all basenames or replaces
+        x side by side words with y words (default=1 for simple 1 word substitution).
         Additionally, a plugin name and file extension can be specified to filter on."""
 
+        if amount:
+            x = amount
+            y = amount
```
benoitryder: If there are only 2 values to configure, don't use 3 parameters, or properly check for conflicting parameters.

Morilli: Hm, do you think not having the ``amount`` parameter and just x and y is better? I thought having a parameter for both would be comfortable, but it might very well just be unnecessary.

benoitryder: I thinks it's fine to only have `x` and `y`. But we should find better names for these parameters.
Something like `nsrc`/`ndst` or `nfrom`/`nto`.

Morilli: Not sure about the name. Tell me if you would prefer a different name.

PR #16, opened 2019-05-15
Add SKN as part of the extracted files

stelar7:
Im sorry about the quotes, [BLACK](https://github.com/python/black) did that by itself T_T

### DIFF
--- a/cdragontoolbox/export.py
+++ b/cdragontoolbox/export.py
@@ -10,7 +10,7 @@ from PIL import Image
 from .storage import PatchVersion
 from .wad import Wad
 from .binfile import BinFile
-from .sknfile import SKNFile
+from .sknfile import SknFile
 from .tools import write_file_or_remove
 
 logger = logging.getLogger(__name__)
@@ -399,7 +399,7 @@ class CdragonRawPatchExporter:
         exporter.converters = [
             ImageConverter(('.dds', '.tga')),
             BinConverter(re.compile(r'^game/data/characters/[^/.]*/(?:skins/)?[^/.]*\.bin$')),
-            SKNConverter([".skn"]),
+            SknConverter([".skn"]),
         ]
         exporter.add_patch_files(patch)
         return exporter
@@ -551,7 +551,7 @@ class BinConverter(FileConverter):
         binfile = BinFile(fin)
         fout.write(json.dumps(binfile.to_serializable()).encode('ascii'))
 
-class SKNConverter(FileConverter):
+class SknConverter(FileConverter):
     def __init__(self, extensions):
         self.extensions = extensions
 
--- a/cdragontoolbox/export.py
+++ b/cdragontoolbox/export.py
@@ -558,7 +558,7 @@ class SknConverter(FileConverter):
 
     def handle_path(self, path):
         if self.extensions:
-            base, ext = os.path.splitext(path)
+            _, ext = os.path.splitext(path)
             if ext in self.extensions:
                 return path
         return None
@@ -568,10 +568,10 @@ class SknConverter(FileConverter):
         fout.close()
         os.remove(fout.name)
         # create a new folder with the same name as the export file, that we use to export the obj files to
-        base, ext = os.path.splitext(fout.name)
+        base, _ = os.path.splitext(fout.name)
         os.makedirs(base, exist_ok=True)
 
-        sknfile = SKNFile(fin)
+        sknfile = SknFile(fin)
         for entry in sknfile.entries:
             name = os.path.join(base, entry["name"] + ".obj")
             with open(name, "w") as f:
--- a/cdragontoolbox/export.py
+++ b/cdragontoolbox/export.py
@@ -10,6 +10,7 @@ from PIL import Image
 from .storage import PatchVersion
 from .wad import Wad
 from .binfile import BinFile
+from .sknfile import SKNFile
 from .tools import write_file_or_remove
 
 logger = logging.getLogger(__name__)
@@ -398,6 +399,7 @@ class CdragonRawPatchExporter:
         exporter.converters = [
             ImageConverter(('.dds', '.tga')),
             BinConverter(re.compile(r'^game/data/characters/[^/.]*/(?:skins/)?[^/.]*\.bin$')),
+            SKNConverter([".skn"]),
         ]
         exporter.add_patch_files(patch)
         return exporter
@@ -549,3 +551,28 @@ class BinConverter(FileConverter):
         binfile = BinFile(fin)
         fout.write(json.dumps(binfile.to_serializable()).encode('ascii'))
 
+class SKNConverter(FileConverter):
+    def __init__(self, extensions):
+        self.extensions = extensions
+
+
+    def handle_path(self, path):
+        if self.extensions:
+            base, ext = os.path.splitext(path)
+            if ext in self.extensions:
+                return path
+        return None
+
+    def convert_to_file(self, fin, fout):
+        # close the writer and delete the extracted file, as each skn file has multiple output files
+        fout.close()
+        os.remove(fout.name)
+        # create a new folder with the same name as the export file, that we use to export the obj files to
+        base, ext = os.path.splitext(fout.name)
+        os.makedirs(base, exist_ok=True)
+
+        sknfile = SKNFile(fin)
+        for entry in sknfile.entries:
+            name = os.path.join(base, entry["name"] + ".obj")
+            with open(name, "w") as f:
+                f.write(sknfile.to_obj(entry))
--- a/cdragontoolbox/sknfile.py
+++ b/cdragontoolbox/sknfile.py
@@ -1,0 +1,71 @@
+from .tools import BinParser
+
+
+class SKNFile:
+    def __init__(self, file):
+        if isinstance(file, str):
+            file = open(file, "rb")
+        if file.read(4) != b"\x33\x22\x11\x00":
+            raise ValueError("missing magic code")
+
+        f = BinParser(file)
+
+        self.major, self.minor, self.count = f.unpack("<HHI")
+        self.entries = [self.read_object(f) for i in range(self.count)]
+
+        if self.major == 4:
+            self.unknown, = f.unpack("<I")
+
+        self.index_count, self.vertex_count = f.unpack("<II")
+
+        if self.major == 4:
+            self.vertex_size, = f.unpack("<I")
+            self.contains_tangent = bool(f.unpack("<I")[0])
+            self.bounding_box_min = f.unpack("<fff")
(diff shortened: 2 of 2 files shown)


benoitryder:
Regarding the auto-formatting, can't you revert those changes to the Git commit?

tests/test_export.py
```diff
@@ -4,18 +4,29 @@
...
+
+def test_wad_export_single_with_transform():
+    exporter = cdtb_export.Exporter("D:/export")
+    exporter.converters = [
+        # those are the converters used on raw
+        cdtb_export.ImageConverter([".dds", ".tga"]),
+        cdtb_export.BinConverter(extensions=[".bin"]),
+        cdtb_export.SKNConverter([".skn"]),
+    ]
+    exporter.add_path("D:/cdragon/Ahri.wad.client", "Ahri")
```
benoitryder: Don't use hardcoded absolute paths in automatic/unitary tests.

cdragontoolbox/export.py
```diff
@@ -536,16 +537,49 @@ def convert_to_file(self, fin, fout):
             # "OSError: cannot identify image file" happen for some files with a wrong extension
             raise FileConversionError("cannot convert image to PNG")
 
+
 class BinConverter(FileConverter):
-    def __init__(self, regex):
+    def __init__(self, regex=None, extensions=None):
```
benoitryder: The `extensions` parameter is not needed; it overlaps with `regex` and its use can be replaced by using an appropriated `regex` value.
You only use it in tests, in which you could use `BinConverter(re.compile(r"\.bin$"))`

stelar7:
There, I think that fixed it.

cdragontoolbox/export.py
```diff
@@ -549,3 +551,28 @@ def convert_to_file(self, fin, fout):
...
+
+    def handle_path(self, path):
+        if self.extensions:
+            base, ext = os.path.splitext(path)
+            if ext in self.extensions:
+                return path
+        return None
+
+    def convert_to_file(self, fin, fout):
+        # close the writer and delete the extracted file, as each skn file has multiple output files
```
benoitryder: It's important that actual output files are known and returned by `handle_path()`.

The current `FileConverter` does not support one file to multiple files conversions (because it was not needed).
If it is now needed it should be extended to support it properly. Note that current approach allowed to avoid recomputing the converted output path by storing it allong with the converter; this will need to be changed.

I'll think about it and try to find a good solution to this.

cdragontoolbox/sknfile.py
```diff
@@ -0,0 +1,71 @@
...
+class SKNFile:
+    def __init__(self, file):
+        if isinstance(file, str):
+            file = open(file, "rb")
+        if file.read(4) != b"\x33\x22\x11\x00":
+            raise ValueError("missing magic code")
+
+        f = BinParser(file)
+
+        self.major, self.minor, self.count = f.unpack("<HHI")
```
benoitryder: `count` is just a temporary and does not need to be stored on `self`.
If you need the entry count afterwards, you can already use `len(self.entries)`.

cdragontoolbox/sknfile.py
```diff
@@ -0,0 +1,71 @@
...
+
+        if self.major == 4:
+            self.vertex_size, = f.unpack("<I")
+            self.contains_tangent = bool(f.unpack("<I")[0])
+            self.bounding_box_min = f.unpack("<fff")
+            self.bounding_box_max = f.unpack("<fff")
+            self.bounding_sphere_location = f.unpack("<fff")
+            self.bounding_sphere_radius, = f.unpack("<f")
+
+        self.indecies = [f.unpack("<H")[0] for i in range(self.index_count)]
```
benoitryder: The plural of `index` is `indices`, not `indecies`.

cdragontoolbox/sknfile.py
```diff
@@ -0,0 +1,71 @@
...
+        self.vertices = [self.read_vertex(f) for i in range(self.vertex_count)]
+
+        for entry in self.entries:
+            entry["vertices"] = self.vertices[entry["start_vertex"] : entry["start_vertex"] + entry["vertex_count"]]
+            entry["indecies"] = self.indecies[entry["start_index"] : entry["start_index"] + entry["index_count"]]
+            entry["indecies"] = [x - (0 if x > entry["start_vertex"] else entry["start_vertex"]) for x in entry["indecies"]]
+
```
benoitryder: - `vertices` and `indecies` are temporaries
- You should not set then override `entry["indecies"]`. Just use temporary and only set the final, true value.
- You don't need `start_index`, `start_vertex`, etc. in the final entries. You should parse them in `read_object()` but keep on `self` entries with only the relevant, non-redundant fields; something like this:
```py
entries_info = [self.read_object(f) for i in range(self.count)]
...
self.entries = []
for info in entries_info:
    self.entries.append({"name": info["name"], "vertices": ..., ", "indices": ...})
```

cdragontoolbox/sknfile.py
```diff
@@ -0,0 +1,71 @@
...
+            content += "vt %s %s\n" % vert["uv"]
+            content += "vn %s %s %s\n" % vert["normal"]
+
+        i = 0
+        while i < entry["index_count"]:
+            a = entry["indecies"][i] + 1
+            b = entry["indecies"][i + 1] + 1
+            c = entry["indecies"][i + 2] + 1
+            content += "f {0}/{0}/{0} {1}/{1}/{1}/ {2}/{2}/{2}\n".format(a, b, c)
+            i += 3
```
benoitryder: You can use `range()` to iterate 3 by 3 and simplify a bit the unpacking and the `+1`:
```py
for i in range(0, len(entry["indices"]), 3):
    a, b, c = entry["indices"][i:i+3]
    content += "f {0}/{0}/{0} {1}/{1}/{1}/ {2}/{2}/{2}\n".format(a + 1, b + 1, c + 1)
```

cdragontoolbox/export.py
```diff
@@ -549,3 +551,28 @@ def convert_to_file(self, fin, fout):
         binfile = BinFile(fin)
         fout.write(json.dumps(binfile.to_serializable()).encode('ascii'))
 
+class SKNConverter(FileConverter):
```
benoitryder: Use `SknConverter` rather than `SKNConverter`.

cdragontoolbox/sknfile.py
```diff
@@ -0,0 +1,71 @@
+from .tools import BinParser
+
+
+class SKNFile:
```
benoitryder: Use `SknFile` rather than `SKNFile`.

cdragontoolbox/export.py
benoitryder: Actually, since you put every file in a single directory, that might work without too much work. This would still require some changes to handle the fact that extracted "files" can now be directories (for instance, in `clean_output_dir()`).

tests/test_export.py
stelar7: the test is removed from the updated version.

cdragontoolbox/export.py
stelar7: Removed from the updated version.

cdragontoolbox/export.py
stelar7: I'm leaving this, as I'm not sure what changes you'd want.

stelar7:
Untested changes, but it shuold be ok..?

cdragontoolbox/export.py
stelar7: If/when we end up exporting WPK and BNK files (voicelines and sfx), they also have multiple output files

benoitryder: Would this be multiples files grouped in a directory or multiple files in the same (parent) directory (as it is for skn)?

cdragontoolbox/sknfile.py
```diff
@@ -0,0 +1,83 @@
...
+            self.entries = [{"name": "Unknown", "vertices": vertices, "indices": [(x + 1) for x in indices]}]
+            return
+
+        count, = f.unpack("<I")
+        self.entries = [self.read_object(f) for i in range(count)]
+
+        if self.major == 4:
+            self.unknown, = f.unpack("<I")
+
+        self.index_count, self.vertex_count = f.unpack("<II")
```
benoitryder: You don't need to store those two values on `self`.

cdragontoolbox/sknfile.py
```diff
@@ -0,0 +1,83 @@
...
+            entry["vertices"] = vertices[entry["start_vertex"] : entry["start_vertex"] + entry["vertex_count"]]
+            entry["indices"] = [
+                (x + 1) - (0 if x < entry["start_vertex"] else entry["start_vertex"]) for x in indices[entry["start_index"] : entry["start_index"] + entry["index_count"]]
+            ]
+            
+            # remove redundant information
+            entry.pop("start_vertex", None)
+            entry.pop("start_index", None)
+            entry.pop("vertex_count", None)
+            entry.pop("index_count", None)
```
benoitryder: You can simply use `del entry["start_vertex"]`. `pop()` is meant to remove and return the value. Here you don't need it.

cdragontoolbox/export.py
stelar7: they should act the same as skn files with multiple models.

cdragontoolbox/sknfile.py
```diff
@@ -0,0 +1,84 @@
...
+    def to_obj(self, entry) -> str:
+        content = ""
+        for vert in entry["vertices"]:
+            content += "v %s %s %s\n" % vert["position"]
+            content += "vt %s %s\n" % vert["uv"]
+            content += "vn %s %s %s\n" % vert["normal"]
+
+        for i in range(0, len(entry["indices"]), 3):
+            a, b, c = entry["indices"][i : i + 3]
+            content += "f {0}/{0}/{0} {1}/{1}/{1}/ {2}/{2}/{2}\n".format(a, b, c)
```
Morilli: From my testing, ``content += "".join(["f {0}/{0}/{0} {1}/{1}/{1}/ {2}/{2}/{2}\n".format(*(entry["indices"][i : i + 3])) for i in range(0, len(entry["indices"]), 3)])`` is faster than this for loop. Can probably be argued whether or not the speed is relevant / readability is affected.

stelar7: I'd say we prefer readability over performance here. Unless there is a _massive_ gain

benoitryder: I agree with stelar7. The difference should not be relevant compared to other processing in this case.

stelar7:
closing as this was added in 1a08aa29c792c50a4474c8232845ac86733023ba

PR #19, opened 2019-07-12
Add a truncation method for extract

Morilli:
Made mainly because of the ``data/skins_skinxx_skins_skinxx.....bin`` files which could become so large they can't be fully extracted on a normal file system because of the 255 character component limit. Open for suggestions regarding code style and general functionality.

### DIFF
--- a/cdragontoolbox/wad.py
+++ b/cdragontoolbox/wad.py
@@ -232,7 +232,7 @@ class Wad:
                     continue
 
                 basename, ext = os.path.splitext(filename)
-                wadfile.path = os.path.join(path, f"{basename[:255-21]}.{hex(wadfile.path_hash)[2:]}{ext}")
+                wadfile.path = os.path.join(path, f"{basename[:255-17-len(ext)]}.{wadfile.path_hash:016x}{ext}")
 
     def extract(self, output, overwrite=True):
         """Extract WAD file
--- a/cdragontoolbox/wad.py
+++ b/cdragontoolbox/wad.py
@@ -222,6 +222,17 @@ class Wad:
                 else:
                     wadfile.path = f"{path}/{wadfile.path_hash:016x}"
 
+    def sanitize_paths(self):
+        """Truncate files that have a length of over 255"""
+
+        for wadfile in self.files:
+            path, filename = os.path.split(wadfile.path)
+            if (len(filename) < 256):
+                continue
+
+            basename, ext = os.path.splitext(filename)
+            wadfile.path = os.path.join(path, basename[:255-21] + "." + hex(wadfile.path_hash)[2:] + ext)
+
     def extract(self, output, overwrite=True):
         """Extract WAD file
 
@@ -231,6 +242,7 @@ class Wad:
         logger.info(f"extracting {self.path} to {output}")
 
         self.set_unknown_paths("unknown")
+        self.sanitize_paths()
 
         with open(self.path, 'rb') as fwad:
             for wadfile in self.files:
--- a/cdragontoolbox/wad.py
+++ b/cdragontoolbox/wad.py
@@ -223,15 +223,16 @@ class Wad:
                     wadfile.path = f"{path}/{wadfile.path_hash:016x}"
 
     def sanitize_paths(self):
-        """Truncate files that have a length of over 255"""
+        """Truncate files whose basename has a length of at least 255"""
 
         for wadfile in self.files:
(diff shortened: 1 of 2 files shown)


cdragontoolbox/wad.py
```diff
@@ -222,6 +222,17 @@ def set_unknown_paths(self, path):
                 else:
                     wadfile.path = f"{path}/{wadfile.path_hash:016x}"
 
+    def sanitize_paths(self):
+        """Truncate files that have a length of over 255"""
+
+        for wadfile in self.files:
+            path, filename = os.path.split(wadfile.path)
+            if (len(filename) < 256):
```
benoitryder: No need for brackets around `if` conditions in Python.

cdragontoolbox/wad.py
```diff
@@ -222,6 +222,17 @@ def set_unknown_paths(self, path):
...
+                continue
+
+            basename, ext = os.path.splitext(filename)
+            wadfile.path = os.path.join(path, basename[:255-21] + "." + hex(wadfile.path_hash)[2:] + ext)
```
benoitryder: You can use f-strings for the formatting.
Something like that: `f"{basename[:255-21}}.{wadfile.path_hash:016x}{ext}`

cdragontoolbox/wad.py
```diff
@@ -222,6 +222,17 @@ def set_unknown_paths(self, path):
                 else:
                     wadfile.path = f"{path}/{wadfile.path_hash:016x}"
 
+    def sanitize_paths(self):
+        """Truncate files that have a length of over 255"""
```
benoitryder: I think you should also truncate basename whose length is exactly 255.
This way, all files whose basename is 255 use the "truncated format". This could be use to detect truncated paths.
I'm not sure this will be used one day, but it costs nothing. :)

cdragontoolbox/wad.py
```diff
@@ -222,6 +222,17 @@ def set_unknown_paths(self, path):
...
+
+        for wadfile in self.files:
+            path, filename = os.path.split(wadfile.path)
```
benoitryder: You should check that `wadfile.path is not None`. `sanitize_path()` could have been called on a wad with unresolved paths.

cdragontoolbox/wad.py
```diff
@@ -231,6 +242,7 @@ def extract(self, output, overwrite=True):
         logger.info(f"extracting {self.path} to {output}")
 
         self.set_unknown_paths("unknown")
+        self.sanitize_paths()
```
benoitryder: Exporters don't call `extract()` directly so they won't use sanitized paths.
You probably want to call `sanitize_path()` in `Exporter.add_path()` (after call to `wad.guess_extensions()`).

cdragontoolbox/wad.py
Morilli: Also not sure whether that should be done, but if you think it's better that way, I'll add it.

Morilli:
Should have resolved all issues. I'm only unsure about ``if wadfile.path is not None`` ?= ``if wadfile.path`` but I believe it should still be correct.

cdragontoolbox/wad.py
```diff
@@ -222,6 +222,18 @@ def set_unknown_paths(self, path):
...
+        """Truncate files whose basename has a length of at least 255"""
+
+        for wadfile in self.files:
+            if wadfile.path:
+                path, filename = os.path.split(wadfile.path)
+                if len(filename) < 255:
+                    continue
+
+                basename, ext = os.path.splitext(filename)
+                wadfile.path = os.path.join(path, f"{basename[:255-21]}.{hex(wadfile.path_hash)[2:]}{ext}")
```
benoitryder: - I didn't realized during the first path, but the lenght from the original `basename` should depend on `ext`' lengths.
- Don't use `{hex(...)}`, use `{value:016x}` to format `value` as hex (16 digits, with leading 0s).

Morilli: > but the lenght from the original basename should depend on ext' lengths

That's true, but I'm pretty much assuming I will always have a bin file here, so I thought it was fine as-is.

benoitryder: It's usually better to adopt a generic solution from the start when it does not cost much more (as here).

Morilli: Okay, should be resolved now.

PR #20, opened 2019-07-19
Add an skn extract method

Morilli:
Simple method to extract a .skn file, just like it's done in ``export``. Was thinking about adding a default output path, but as I'm not sure about that I've left it out for now.

### DIFF
--- a/cdragontoolbox/__main__.py
+++ b/cdragontoolbox/__main__.py
@@ -285,9 +285,11 @@ def command_export(parser, args):
 
 def command_skn_extract(parser, args):
     if not os.path.isfile(args.skn):
-        parser.error("SKN file does not exist")
+        parser.error(f"SKN file not found: {args.skn}")
 
     sknfile = SknFile(args.skn)
+    if args.output is None:
+        args.output = os.path.splitext(args.skn)[0]
     os.makedirs(args.output, exist_ok=True)
     for entry in sknfile.entries:
         name = os.path.join(args.output, entry["name"] + ".obj")
@@ -297,7 +299,7 @@ def command_skn_extract(parser, args):
 
 def command_bin_dump(parser, args):
     if not os.path.isfile(args.bin):
-        parser.error("BIN file does not exist")
+        parser.error(f"BIN file not found: {args.bin}")
 
     with open(args.bin, 'rb') as f:
         binfile = BinFile(f)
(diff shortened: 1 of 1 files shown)


cdragontoolbox/__main__.py
```diff
@@ -282,6 +283,18 @@ def command_export(parser, args):
         exporter.process(overwrite=overwrite)
 
 
+def command_skn_extract(parser, args):
+    if not os.path.isfile(args.skn):
+        parser.error("SKN file does not exist")
```
benoitryder: `does not exist` -> `not found`

Also, you could add the file path: `f"SKN file not found: {args.skn}"`

cdragontoolbox/__main__.py
```diff
@@ -429,6 +442,14 @@ def create_parser():
     subparser.add_argument('bin',
                            help="BIN file to extract")
 
+    # skn files commands
+
+    subparser = subparsers.add_parser('skn-extract',
+                                      help="Extract an SKN file to a folder")
```
benoitryder: For consistency with other commands:
- `extract` (lowercase)
- `folder` -> `directory`

cdragontoolbox/__main__.py
```diff
@@ -429,6 +442,14 @@ def create_parser():
...
+    subparser.add_argument('skn',
+                           help="SKN file to extract")
+    subparser.add_argument('output',
+                           help="output folder")
```
benoitryder: I think it would be better to use a `-o`/`--output` option, as elsewhere.
You could make it default (if `None`/unset) to a directory based on `skn` value.

Also, `folder` -> `directory`.

cdragontoolbox/__main__.py
Morilli: For this I just copied what was done in ``bin_dump``. Should I change the code there as well, to keep it consistent?

cdragontoolbox/__main__.py
Morilli: If you're fine with a (unspecified) default value, I'll gladly add that. This will help my personal use case as well.

cdragontoolbox/__main__.py
benoitryder: Yes.

cdragontoolbox/__main__.py
benoitryder: Yes, it's fine. Juste use `None` as default value and convert it to the final value later, based on `args.skn`.

Morilli:
Should be good to go from my side then.

PR #21, opened 2019-08-16
Fix regexes, add tiers bin paths

Morilli:


cdragontoolbox/hashes.py
```diff
@@ -722,14 +724,14 @@ def grep_wad(self, wad):
...
-                if wadfile.ext in ('dds', 'jpg', 'png', 'tga', 'ttf', 'otf', 'ogg', 'webm','anm',
+                if wadfile.ext in ('dds', 'jpg', 'png', 'tga', 'ttf', 'otf', 'ogg', 'webm', 'anm',
                                    'skl', 'skn', 'scb', 'sco', 'troybin', 'luabin', 'luabin64', 'bnk', 'wpk'):
                     continue # don't grep filetypes known to not contain full paths
 
                 data = wadfile.read_data(f)
                 if wadfile.ext in ('bin', 'inibin'):
                     # bin files: find strings based on prefix, then parse the length
-                    for m in re.finditer(br'(..)((?:ASSETS|DATA|Characters|Shaders)/[0-9a-zA-Z_. /-]+)', data):
+                    for m in re.finditer(br'(?=(..)((?:ASSETS|DATA|Characters|Shaders)/[0-9a-zA-Z_. /-]+))', data):
```
benoitryder: Which case do you want to address with this change?

Morilli: This previously consumed the ``(..)``, which in case of linked files in the header of bin files could result in the string length for the next string not being found and thereby resulting in the linked string not being guessed.

benoitryder: If we use a look-ahead assertion, then I think we can speed up the process a bit by not matching the length, or even the whole string.

The following code should find all paths with a known prefix and be faster: no backtracking, no need to match the "suffix".
Not that it's incomplete (see comments), but it gives the idea.
```py
for m in re.finditer(br'(?:ASSETS|DATA|Characters|Shaders)/', data):
    i = m.start(0)
    # Note: this check could be avoided by finding in `data[2:]` or using a pattern object which support a `pos` parameter
    if i < 0:
        continue
    n = data[i-2] + (data[i-1] << 8)
    # Note: this is incomplete: if the string contains non-ASCII characters, this will fail
    # We could stop at the first non-printable character, or decode as  `latin1`
    path = data[i:i+n].lower().decode('ascii')
```

PR #23, opened 2019-08-26
Add TFT data export script

stelar7:
I'd prefer it if this was ran on raw so its always up-to-date

### DIFF
--- a/cdragontoolbox/binfile.py
+++ b/cdragontoolbox/binfile.py
@@ -134,15 +134,6 @@ class BinObjectWithFields:
         except KeyError:
             return default
 
-    def getr(self, key, default=None):
-        for v in self.fields:
-            if v.name.hex() == key:
-                return v
-        return default
-
-    def getrv(self, key, default=None):
-        return self.getr(key, default).value
-
     def to_serializable(self):
         return dict(f.to_serializable() for f in self.fields)
 
--- a/cdragontoolbox/binfile.py
+++ b/cdragontoolbox/binfile.py
@@ -51,13 +51,16 @@ class BinHashBase:
     def __str__(self):
         if self.s is not None:
             return self.s
-        return f"{{{self.h:08x}}}"
+        return f"{{{self.hex()}}}"
 
     __repr__ = __str__
     to_serializable = __str__
 
     def __hash__(self):
         return self.h
+
+    def hex(self):
+        return f"{self.h:08x}"
 
 class BinHashValue(BinHashBase):
     """Hashed name in bin files (hash type)"""
@@ -67,7 +70,7 @@ class BinHashValue(BinHashBase):
     def __repr__(self):
         if self.s is not None:
             return repr(self.s)
-        return f"{{{self.h:08x}}}"
+        return f"{{{self.hex()}}}"
 
 class BinEntryPath(BinHashBase):
     """Path of a bin entry (top level element)"""
@@ -77,7 +80,7 @@ class BinEntryPath(BinHashBase):
     def __repr__(self):
         if self.s is not None:
             return repr(self.s)
-        return f"{{{self.h:08x}}}"
+        return f"{{{self.hex()}}}"
 
 class BinFieldName(BinHashBase):
     """Name of a struct field"""
@@ -130,6 +133,15 @@ class BinObjectWithFields:
             return self[key].value
         except KeyError:
             return default
+
+    def getr(self, key, default=None):
+        for v in self.fields:
+            if v.name.hex() == key:
+                return v
+        return default
+
+    def getrv(self, key, default=None):
+        return self.getr(key, default).value
 
     def to_serializable(self):
         return dict(f.to_serializable() for f in self.fields)
--- a/tft.py
+++ b/tft.py
@@ -1,7 +1,6 @@
+import argparse
 import json
 import glob
-import mmap
-import copy
 import re
 import os
 from json import JSONEncoder
@@ -17,7 +16,7 @@ class NaiveJsonEncoder(JSONEncoder):
         return other.__dict__
 
 
-class TFTTransformer:
+class TftTransformer:
     def parse(self, input):
         map22_file = os.path.join(input, "data", "maps", "shipping", "map22", "map22.bin")
         character_folder = os.path.join(input, "data", "characters")
@@ -33,7 +32,7 @@ class TFTTransformer:
 
         self.data = {"items": items, "traits": traits, "champs": champs}
 
-    def export(self, output, font_files=[], output_template=False):
+    def export(self, output, font_files=None, output_template=False):
         os.makedirs(output, exist_ok=True)
         template = json.dumps(self.data, cls=NaiveJsonEncoder, indent=4, sort_keys=True)
 
@@ -41,11 +40,14 @@ class TFTTransformer:
             with open(os.path.join(output, "template.json"), "w", encoding="utf-8") as f:
                 f.write(template)
 
+        if not font_files:
+            return
+
         for language in font_files:
             use = template
             code = os.path.basename(language)[11:-4]
 
-            with open(language, "r", encoding="utf-8") as f:
+            with open(language, encoding="utf-8") as f:
                 for line in f:
                     if line.startswith("tr"):
                         key, val = line.split("=", 1)
@@ -71,29 +73,27 @@ class TFTTransformer:
 
             items.append(
                 {
-                    "internal": {"hex": item.path.hex(), "name": item.getv("mName")},
+                    "internal": {"hex": item.path.h, "name": item.getv("mName")},
                     "id": item.getv("mId"),
-                    "name": item.getrv("c3143d66"),
-                    "desc": item.getrv("765f18da"),
+                    "name": item.getv(0xC3143D66),
+                    "desc": item.getv(0x765F18DA),
                     "icon": item.getv("mIconPath"),
-                    "from": [x.hex() for x in item.getrv("8b83ba8a")] if item.getr("8b83ba8a") else [],
+                    "from": [x.h for x in item.getv(0x8B83BA8A, [])],
                     "effects": effects,
                 }
             )
 
-        # replace the hex key gotten earlier with the id
+        item_ids = {x["internal"]["hex"]: x["id"] for x in items}
         for item in items:
-            # there must be some list comprehension way of doing this...
-            for old in item["from"]:
-                key = [x for x in items if x["internal"]["hex"] == old][0]["id"]
-                item["from"].insert(0, key)
-                item["from"].remove(old)
+            item["from"] = [item_ids[x] for x in item["from"]]
 
         return items
 
     def parse_champs(self, map22, traits, character_folder):
         champ_collection = [x for x in map22.entries if x.type == "TftShopData"]
         champs = []
+
+        trait_names = {x["internal"]["hash"]: x["internal"]["name"] for x in traits}
 
         for champ in champ_collection:
             name = champ.getv("mName")
@@ -104,16 +104,14 @@ class TFTTransformer:
             char_bin = BinFile(os.path.join(character_folder, name[4:], name[4:] + ".bin"))
             champ_id = [x.getv("characterToolData").getv("championId") for x in char_bin.entries if x.type == "CharacterRecord"][0]
 
-            champ_traits = []
             tft_bin = BinFile(os.path.join(character_folder, name, name + ".bin"))
             record = [x for x in tft_bin.entries if x.type == "TFTCharacterRecord"][0]
-            for trait in record.getv("mLinkedTraits"):
-                champ_traits.append([x["internal"]["name"] for x in traits if x["internal"]["hash"] == trait.hex()][0])
+            champ_traits = [trait_names[trait.h] for trait in record.getv("mLinkedTraits", [])]
 
             stats_obj = {
                 "hp": record.getv("baseHP"),
-                "mana": record.getv("primaryAbilityResource").getv("arBase"),
-                "initialMana": record.getv("mInitialMana") or 0,
+                "mana": record.getv("primaryAbilityResource").getv("arBase", 100),
+                "initialMana": record.getv("mInitialMana", 0),
                 "damage": record.getv("BaseDamage"),
                 "armor": record.getv("baseArmor"),
                 "magicResist": record.getv("baseSpellBlock"),
@@ -132,13 +130,13 @@ class TFTTransformer:
             for value in ability.getv("mDataValues"):
                 variables.append({"name": value.getv("mName"), "value": value.getv("mValues")})
 
-            ability_obj = {"name": champ.getrv("87a69a5e"), "desc": champ.getrv("bc4f18b3"), "icon": champ.getv("mPortraitIconPath"), "variables": variables}
+            ability_obj = {"name": champ.getv(0x87A69A5E), "desc": champ.getv(0xBC4F18B3), "icon": champ.getv("mPortraitIconPath"), "variables": variables}
 
             champs.append(
                 {
                     "id": champ_id,
-                    "name": champ.getrv("c3143d66"),
-                    "cost": champ.getv("mRarity") or 1,
+                    "name": champ.getv(0xC3143D66),
+                    "cost": champ.getv("mRarity", 1),
                     "icon": champ.getv("mIconPath"),
                     "traits": champ_traits,
                     "stats": stats_obj,
@@ -166,9 +164,9 @@ class TFTTransformer:
 
             traits.append(
                 {
-                    "internal": {"hash": trait.path.hex(), "name": trait.getv("mName")},
-                    "name": trait.getrv("c3143d66"),
-                    "desc": trait.getrv("765f18da"),
+                    "internal": {"hash": trait.path.h, "name": trait.getv("mName")},
+                    "name": trait.getv(0xC3143D66),
+                    "desc": trait.getv(0x765F18DA),
                     "icon": trait.getv("mIconPath"),
                     "effects": effects,
                 }
@@ -178,12 +176,16 @@ class TFTTransformer:
 
 
 if __name__ == "__main__":
-    input_folder = os.path.join("D:", "pbe")
-    output_folder = os.path.join("D:", "pbe", "tft_temp")
+    parser = argparse.ArgumentParser()
+    parser.add_argument("input", help="directory with extracted bin files")
+    parser.add_argument("-o", "--output", default="tft", help="output directory")
+    args = parser.parse_args()
 
-    font_folder = os.path.join(input_folder, "data", "menu")
+    font_folder = os.path.join(args.input, "data", "menu")
     lang_files = glob.glob(f"{font_folder}/fontconfig_*_*.txt")
 
-    tft_transformer = TFTTransformer()
-    tft_transformer.parse(input_folder)
-    tft_transformer.export(output_folder, lang_files, True)
+    tft_transformer = TftTransformer()
+    tft_transformer.parse(args.input)
+    tft_transformer.export(args.output, lang_files, True)
+
+    #py tft.py -o D:/tft/ D:/pbe/
--- a/tft.py
+++ b/tft.py
@@ -1,0 +1,189 @@
+import json
+import glob
+import mmap
+import copy
+import re
+import os
+from json import JSONEncoder
+from cdragontoolbox.binfile import BinFile, BinHashBase, BinHashValue
+
+
+class NaiveJsonEncoder(JSONEncoder):
+    def default(self, other):  # pylint: disable=method-hidden
+        if isinstance(other, BinHashBase):
+            if other.s is None:
+                return other.hex()
+            return other.s
+        return other.__dict__
+
+
+class TFTTransformer:
+    def parse(self, input):
+        map22_file = os.path.join(input, "data", "maps", "shipping", "map22", "map22.bin")
+        character_folder = os.path.join(input, "data", "characters")
+
+        map22 = BinFile(map22_file)
+        items = self.parse_items(map22)
+        traits = self.parse_traits(map22)
+        champs = self.parse_champs(map22, traits, character_folder)
+
+        # clean up data for JSON export
+        [x.pop("internal") for x in traits]
+        [x.pop("internal") for x in items]
+
+        self.data = {"items": items, "traits": traits, "champs": champs}
+
+    def export(self, output, font_files=[], output_template=False):
+        os.makedirs(output, exist_ok=True)
+        template = json.dumps(self.data, cls=NaiveJsonEncoder, indent=4, sort_keys=True)
+
+        if output_template:
+            with open(os.path.join(output, "template.json"), "w", encoding="utf-8") as f:
+                f.write(template)
+
+        for language in font_files:
+            use = template
+            code = os.path.basename(language)[11:-4]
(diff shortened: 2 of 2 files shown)


cdragontoolbox/binfile.py
```diff
@@ -36,6 +36,7 @@ class BinHashBase:
     def __init__(self, h):
         self.h = h
         self.s = self.hashfile.load().get(h)
+        self.hex = f"{self.h:08x}"
```
benoitryder: There is no need to "cache" the hex representation on the instance.

cdragontoolbox/binfile.py
```diff
@@ -51,7 +52,7 @@ def __eq__(self, other):
     def __str__(self):
         if self.s is not None:
             return self.s
-        return f"{{{self.h:08x}}}"
+        return self.hex
```
benoitryder: This changes the output (it removes surrounding braces), which is already used by other parts of the code.

cdragontoolbox/binfile.py
```diff
@@ -131,6 +132,21 @@ def getv(self, key, default=None):
         except KeyError:
             return default
 
+    def hasr(self, key):
```
benoitryder: What does `hasr` means? (same for `getr` and `getrv`)

tft.py
```diff
@@ -0,0 +1,213 @@
+from cdragontoolbox.binfile import BinFile, BinHashBase, BinHashValue
+from json import JSONEncoder
+import json
+import mmap
+import copy
+import re
+import os
```
benoitryder: Imports are preferably put in this order: standard lib, external libs (dependencies), current package.

tft.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+class Transformer:
+
+    output_prefix = r"D:\pbe\tft_temp"
+    prefix = r"D:\pbe"
+    map22_file = os.path.join(prefix, r"data\maps\shipping\map22\map22.bin")
+    font_folder = os.path.join(prefix, r"data\menu")
+    character_folder = os.path.join(prefix, r"data\characters")
+    exported_file = os.path.join(prefix, r"files.exported.txt")
```
benoitryder: - Don't put hardcoded paths
- Use forward slashes as path delimiters (they are supported on both Windows on Linux so it's ok).

tft.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def __init__(self):
+        with open(self.exported_file, "rb", 0) as f:
+            with mmap.mmap(f.fileno(), 0, access=mmap.ACCESS_READ) as s:
+                self.description_files = re.findall(br"game/data/menu/(fontconfig_.._..\.txt)", s)
```
benoitryder: You could just "glob" `f"{font_folder}/game/data/menu/fontconfig_*_*.txt` (check the `glob` module).

tft.py
```diff
@@ -0,0 +1,213 @@
...
+        self.traits = self.parse_traits(map22)
+        self.champs = self.parse_champs(map22, self.traits)
+        self.items = self.parse_items(map22)
+        lang_files = self.description_files
+
+        # clean up data for JSON export
+        [x.pop("internal") for x in self.traits]
+        [self.items[x].pop("internal") for x in self.items]
+        [self.items[x].pop("newf") for x in self.items]
+        del self.description_files
```
benoitryder: No need to store temporary variables on `self`.

tft.py
```diff
@@ -0,0 +1,213 @@
...
+
+        os.mkdir(self.output_prefix)
```
benoitryder: Prefer `os.makedirs(self.output_prefix, exist_ok=False)` (it will create parent directories and won't fail if the output directory already exists).

tft.py
```diff
@@ -0,0 +1,213 @@
...
+        with open(os.path.join(self.output_prefix, "template.json"), "w+", encoding="utf-8") as f:
+            f.write(template)
+
+        for language in lang_files:
+            use = copy.deepcopy(template)
+            code = str(language, "utf-8")[11:-4]
+
+            font_file = os.path.join(self.font_folder, str(language, "utf-8"))
+            with open(font_file, "rb", 0) as f:
+                with mmap.mmap(f.fileno(), 0, access=mmap.ACCESS_READ) as s:
```
benoitryder: - You won't need all those manual encoding conversions if you use glob and `re.match()` the results to extract the lang.
- No need to use `mmap`, just iterate on lines (see below).
- Don't open the file as binary if it contains text.

```py
with open(font_file) as f:
    for line in f:
        if line.startswith("tr"):
            ....
```

tft.py
```diff
@@ -0,0 +1,213 @@
...
+        [self.items[x].pop("newf") for x in self.items]
+        del self.description_files
+
+        os.mkdir(self.output_prefix)
+        template = json.dumps(self, cls=NaiveJsonEncoder, indent=4, sort_keys=True)
+        with open(os.path.join(self.output_prefix, "template.json"), "w+", encoding="utf-8") as f:
+            f.write(template)
+
+        for language in lang_files:
+            use = copy.deepcopy(template)
```
benoitryder: No need to copy strings, they are immutable.

tft.py
```diff
@@ -0,0 +1,213 @@
...
+            font_file = os.path.join(self.font_folder, str(language, "utf-8"))
+            with open(font_file, "rb", 0) as f:
+                with mmap.mmap(f.fileno(), 0, access=mmap.ACCESS_READ) as s:
+                    for line in iter(s.readline, b""):
+                        line = str(line, "utf-8")
+                        if line.startswith("tr"):
+                            split = line.split("=")
+                            key = split[0][4:-2]
+                            val = split[1][2:-3]
+                            use = use.replace(key, val)
```
benoitryder: If I'm not mistaking, you "blindly" replace strings into JSON encoded data, not checking anything (not even word boundaries). This is definitely not clean.

tft.py
```diff
@@ -0,0 +1,213 @@
...
+            use = copy.deepcopy(template)
+            code = str(language, "utf-8")[11:-4]
+
+            font_file = os.path.join(self.font_folder, str(language, "utf-8"))
+            with open(font_file, "rb", 0) as f:
+                with mmap.mmap(f.fileno(), 0, access=mmap.ACCESS_READ) as s:
+                    for line in iter(s.readline, b""):
+                        line = str(line, "utf-8")
+                        if line.startswith("tr"):
+                            split = line.split("=")
```
benoitryder: - For information, you can expand the split directly
- You should limit to 1 split.

```py
key, val = line.split("=", 1)
key = key[4:-2]
val = val[2:-3]
```

tft.py
```diff
@@ -0,0 +1,213 @@
...
+                            key = split[0][4:-2]
+                            val = split[1][2:-3]
+                            use = use.replace(key, val)
+
+            with open(os.path.join(self.output_prefix, code + ".json"), "w+", encoding="utf-8") as f:
```
benoitryder: Why append? It seems each `code` will be iterated on only once.

tft.py
```diff
@@ -0,0 +1,213 @@
...
+                f.write(use)
+            input()
```
benoitryder: Is it intended to prompt the user?

tft.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def parse_items(self, map22):
+        item_collection = [x for x in map22.entries if x.type == "TftItemData"]
+        items = {}
+
+        for item in item_collection:
+            name = item.getv("mName")
+
+            if name.find("Template") > 0 or name == "TFT_Item_Null":
```
benoitryder: Replace `name.find("Template") > 0` with `"Template" in name`

tft.py
```diff
@@ -0,0 +1,213 @@
...
+            item_obj = {}
+            item_obj["internal"] = {}
+            item_obj["internal"]["hex"] = item.path.hex
+            item_obj["internal"]["name"] = item.getv("mName")
+            item_obj["id"] = item.getv("mId")
+            item_obj["name"] = item.getrv("c3143d66")
+            item_obj["desc"] = item.getrv("765f18da")
+            item_obj["icon"] = item.getv("mIconPath")
+            item_obj["from"] = [x.hex for x in item.getrv("8b83ba8a")] if item.hasr("8b83ba8a") else []
+            item_obj["effects"] = []
```
benoitryder: Prefer use of a literal directly (other occurrences later):
```py
item_obj = {
    "internal": {
        "hex": item.path.hex,
        ....
    },
    ....
}
```

tft.py
```diff
@@ -0,0 +1,213 @@
...
+
+            if not item.get("effectAmounts"):
+                continue
+
+            for effect in item.getv("effectAmounts"):
+                effect_obj = {}
+                effect_obj["name"] = effect.getv("name")
+                effect_obj["value"] = effect.getv("value")
```
benoitryder: Same as above.

tft.py
```diff
@@ -0,0 +1,213 @@
...
+
+                item_obj["effects"].append(effect_obj)
+
+            items[item_obj["id"]] = item_obj
+
+        for item in items:
```
benoitryder: If you need values, iterate on them directly:
```py
for item in items.values():
  # use item instead of items[item]
```

tft.py
```diff
@@ -0,0 +1,213 @@
...
+            # there must be some list comprehention way of doing this...
+
+            items[item]["newf"] = []
+            for old in items[item]["from"]:
+                key = [x for x in items if old in items[x]["internal"]["hex"]][0]
+                items[item]["newf"].append(key)
+
+            items[item]["from"] = items[item]["newf"]
```
benoitryder: Why both `from` and `newf` if they ultimately contain the same value?

tft.py
```diff
@@ -0,0 +1,213 @@
...
+        return items
+
+    def parse_champs(self, map22, traits):
+        champ_collection = [x for x in map22.entries if x.type == "TftShopData"]
+        champs = {}
+
+        for champ in champ_collection:
+            name = champ.getv("mName")
+
+            if (not name.startswith("TFT_")) or name == "TFT_Template":
```
benoitryder: Not need for the brackets.

tft.py
```diff
@@ -0,0 +1,213 @@
...
+
+            champ_obj["ability"] = {}
+            champ_obj["ability"]["name"] = champ.getrv("87a69a5e")
+            champ_obj["ability"]["desc"] = champ.getrv("bc4f18b3")
+            champ_obj["ability"]["icon"] = champ.getv("mPortraitIconPath")
+            champ_obj["ability"]["variables"] = []
+
+            spellName = record.getv("spellNames")[0]
+            if "/" in spellName:
+                spellName = spellName[spellName.find("/") + 1 :]
```
benoitryder: Do you want exactly the 2d part or the last one (the basename)?
If you want the basename, you can use `os.path.basename(spellName)`.

tft.py
```diff
@@ -0,0 +1,213 @@
...
+            champ_obj["stats"]["attackSpeed"] = record.getv("attackSpeed")
+            champ_obj["stats"]["range"] = record.getv("attackRange") // 180
+
+            champ_obj["ability"] = {}
+            champ_obj["ability"]["name"] = champ.getrv("87a69a5e")
+            champ_obj["ability"]["desc"] = champ.getrv("bc4f18b3")
+            champ_obj["ability"]["icon"] = champ.getv("mPortraitIconPath")
+            champ_obj["ability"]["variables"] = []
+
+            spellName = record.getv("spellNames")[0]
```
benoitryder: Use `snake_case` names for variables (`spell_name`).

tft.py
```diff
@@ -0,0 +1,213 @@
...
+                    trait_effects["variables"].append(trait_effect)
+
+            trait_obj["effects"] = trait_effects
+            traits.append(trait_obj)
+
+        return traits
+
+
+if __name__ == "__main__":
+    Transformer()
```
benoitryder: The constructor should not do all the stuff.
You should use `argparse` to parse options (output path, etc.), then (if a class is needed) create an instance with those parameters (creating the instance not do the processing itself), then call a processing method.

cdragontoolbox/binfile.py
stelar7: hasRaw, getRaw, and getRawValue.
I tried using get and getv first, but since they hash the string, it didnt work

tft.py
stelar7: if you dont hardcode paths, how are you supposed to refer to those files?
(atleast the bottom 4, since they are static in cdragon)

tft.py
stelar7: 2nd part

tft.py
stelar7: the keys in this case are unique enough that it works, but if you have a better idea, im all for it

tft.py
stelar7: google said `w+` would create the file if missing, so thats what i went with. if just w is enough, then ill swap it over

cdragontoolbox/binfile.py
```diff
@@ -131,6 +134,15 @@ def getv(self, key, default=None):
         except KeyError:
             return default
 
+    def getr(self, key, default=None):
```
stelar7: used for getting by-hash, since the normal get fails with KeyError if you use a hash

benoitryder: `key_to_hash()` already handles strings, integers (hash values) and `BinHashBase` (wrapped bin values).
This means you can already use `get(0x12345678)` instead of `getr("12345678")` to access fields by hash (same for `getv()` vs. `getrv()`.

You should need to use hex strings for hash values, just use plain integers (as everywhere else) and you'll be fine. (Bonus: it will be faster.)

Also, you can probably use `item[key]` instead of `item.get(key)` most of the time.
The difference is that `[...]` will fail if the key is not found but `get()` will return a default value (`None` by default). If you know the key must exist, `[...]` is shorter.

tft.py
```diff
@@ -0,0 +1,189 @@
...
+class NaiveJsonEncoder(JSONEncoder):
+    def default(self, other):  # pylint: disable=method-hidden
+        if isinstance(other, BinHashBase):
+            if other.s is None:
+                return other.hex()
+            return other.s
+        return other.__dict__
+
+
+class TFTTransformer:
```
benoitryder: Use `TftTransformer` instead of `TFTTransformer`.

tft.py
```diff
@@ -0,0 +1,189 @@
...
+
+        if output_template:
+            with open(os.path.join(output, "template.json"), "w", encoding="utf-8") as f:
+                f.write(template)
+
+        for language in font_files:
+            use = template
+            code = os.path.basename(language)[11:-4]
+
+            with open(language, "r", encoding="utf-8") as f:
```
benoitryder: `"r"` is the default so you don't need to pass it explicitely (and it's usely omitted).

tft.py
```diff
@@ -0,0 +1,189 @@
...
+                }
+            )
+
+        # replace the hex key gotten earlier with the id
+        for item in items:
+            # there must be some list comprehension way of doing this...
+            for old in item["from"]:
+                key = [x for x in items if x["internal"]["hex"] == old][0]["id"]
+                item["from"].insert(0, key)
+                item["from"].remove(old)
```
benoitryder: You cannot reliabily iterate on a list you are updating at the same time. It won't crash but you may obtain unwanted results. I guess you made it work with your `insert()`/`remove()` but I don't think you can safely rely on current behavior.

Also, regarding your comment, you could make `items` a dict for easier lookup. That would also solve the "update and iterate" problem.
```py
items = {}
for item in item_collection:
    ...
    items[item.path.h] = {
        "internal": {"hex": item.path.h, "name": ...},
        ...
    }

# convert "from" values from hex to id
for item in items.values():
    item["from"] = items[item["from"]]["id"]

return items.values()
```

tft.py
```diff
@@ -0,0 +1,189 @@
...
+                effects.append({"name": effect.getv("name"), "value": effect.getv("value")})
+
+            items.append(
+                {
+                    "internal": {"hex": item.path.hex(), "name": item.getv("mName")},
+                    "id": item.getv("mId"),
+                    "name": item.getrv("c3143d66"),
+                    "desc": item.getrv("765f18da"),
+                    "icon": item.getv("mIconPath"),
+                    "from": [x.hex() for x in item.getrv("8b83ba8a")] if item.getr("8b83ba8a") else [],
```
benoitryder: Here you can rely on `get()` to return an empty list as default and avoid the ternary.
And you don't need to convert to a string, just use the integer hash.
```py
[x.h for x in item.getv(0x8b83ba8a, [])]
```

tft.py
```diff
@@ -0,0 +1,189 @@
...
+        traits = self.parse_traits(map22)
+        champs = self.parse_champs(map22, traits, character_folder)
+
+        # clean up data for JSON export
+        [x.pop("internal") for x in traits]
+        [x.pop("internal") for x in items]
+
+        self.data = {"items": items, "traits": traits, "champs": champs}
+
+    def export(self, output, font_files=[], output_template=False):
```
benoitryder: Avoid mutable values are default parameters as they can introduce sneaky bugs if misused.
Default parameters are evaluated once (when the method is parsed), and since lists are used by reference, the same list object will be used for every call.
In your specific case this does not matter since you don't modify `font_files` but it's easy to forget when you change the code so it's considered a good practice to avoid `[]` as default value.

In your case you can simply use `font_files=None` and use `if not font_files: return` before the `for` loop.

tft.py
```diff
@@ -0,0 +1,189 @@
...
+            for trait_set in trait.getv("mTraitSets"):
+                variables = []
+                for effect in trait_set.getv("effectAmounts") if trait_set.get("effectAmounts") else []:
+                    variables.append({"name": effect.getv("name"), "value": effect.getv("value")})
+
+                effects.append({"minUnits": trait_set.getv("mMinUnits"), "maxUnits": trait_set.getv("mMaxUnits"), "variables": variables})
+
+            traits.append(
+                {
+                    "internal": {"hash": trait.path.hex(), "name": trait.getv("mName")},
```
benoitryder: You should use `trait.path` or `trait.path.h` directly instead of the hex string.

tft.py
```diff
@@ -0,0 +1,189 @@
...
+            if "Template" in name or name == "TFT_Item_Null":
+                continue
+
+            effects = []
+            for effect in item.getv("effectAmounts") if item.get("effectAmounts") else []:
+                effects.append({"name": effect.getv("name"), "value": effect.getv("value")})
+
+            items.append(
+                {
+                    "internal": {"hex": item.path.hex(), "name": item.getv("mName")},
```
benoitryder: You should use `item.path` or `item.path.h` directly instead of the hex string.

tft.py
```diff
@@ -0,0 +1,189 @@
...
+            champ_traits = []
+            tft_bin = BinFile(os.path.join(character_folder, name, name + ".bin"))
+            record = [x for x in tft_bin.entries if x.type == "TFTCharacterRecord"][0]
+            for trait in record.getv("mLinkedTraits"):
+                champ_traits.append([x["internal"]["name"] for x in traits if x["internal"]["hash"] == trait.hex()][0])
+
+            stats_obj = {
+                "hp": record.getv("baseHP"),
+                "mana": record.getv("primaryAbilityResource").getv("arBase"),
+                "initialMana": record.getv("mInitialMana") or 0,
```
benoitryder: Use `getv("mInitialMana", 0)`.

tft.py
```diff
@@ -0,0 +1,189 @@
...
+                continue
+
+            char_bin = BinFile(os.path.join(character_folder, name[4:], name[4:] + ".bin"))
+            champ_id = [x.getv("characterToolData").getv("championId") for x in char_bin.entries if x.type == "CharacterRecord"][0]
+
+            champ_traits = []
+            tft_bin = BinFile(os.path.join(character_folder, name, name + ".bin"))
+            record = [x for x in tft_bin.entries if x.type == "TFTCharacterRecord"][0]
+            for trait in record.getv("mLinkedTraits"):
+                champ_traits.append([x["internal"]["name"] for x in traits if x["internal"]["hash"] == trait.hex()][0])
```
benoitryder: Similarly to items, you can convert `traits` to a hash-to-name dict first (at the beginning of the method) and then use regular indexing.
```py
trait_names = {x["internal"]["hash"]: x["internal"]["name"] for x in traits}
...
champ_traits = [trait_names[trait.h] for trait in record["mLinkedTraits"].value]

File: tft.py [hunk 34]
```diff
@@ -0,0 +1,189 @@
...
+                    "effects": effects,
+                }
+            )
+
+        return traits
+
+
+if __name__ == "__main__":
+    input_folder = os.path.join("D:", "pbe")
+    output_folder = os.path.join("D:", "pbe", "tft_temp")
```
benoitryder: Minimal version for arg parsing:
```py
import argparse

parser = argparse.ArgumentParser()
parser.add_argument("input", help="directory with extracted bin files")
parser.add_argument("-o", "--output", default="tft", help="output directory")
args = parser.parse_args()
```
Then use `args.input` and `args.output`.

cdragontoolbox/binfile.py
stelar7: I'd like to keep use hashes since thats what is used elsewhere in cdragon

benoitryder: You can use hashes, just as integers instead of strings. And that would just be in the script: no problem to output them as strings in the resulting JSON.

tft.py
stelar7: `item["from"] is a list, so id still need to .append tho?`

benoitryder: Oh sorry, use: `item["from"] = [items[x]["id"] for x in items["from"]]`

tft.py
```diff
@@ -0,0 +1,189 @@
+import json
+import glob
+import mmap
+import copy
```
benoitryder: Forgot to note that `mmap` and `copy` are not needed anymore.

tft.py
```diff
@@ -0,0 +1,191 @@
...
+        item_collection = [x for x in map22.entries if x.type == "TftItemData"]
+        items = []
+
+        for item in item_collection:
+            name = item.getv("mName")
+            if "Template" in name or name == "TFT_Item_Null":
+                continue
+
+            effects = []
+            for effect in item.getv("effectAmounts") if item.get("effectAmounts") else []:
```
benoitryder: Just use `for effect in item.getv("effectAmounts", []):`

tft.py
```diff
@@ -0,0 +1,191 @@
...
+        traits = []
+
+        for trait in trait_collection:
+            if "Template" in trait.getv("mName"):
+                continue
+
+            effects = []
+            for trait_set in trait.getv("mTraitSets"):
+                variables = []
+                for effect in trait_set.getv("effectAmounts") if trait_set.get("effectAmounts") else []:
```
benoitryder: Same as above: `for effect in trait_set.getv("effectAmounts", []):`

tft.py
```diff
@@ -0,0 +1,191 @@
...
+    args = parser.parse_args()
+
+    font_folder = os.path.join(args.input, "data", "menu")
+    lang_files = glob.glob(f"{font_folder}/fontconfig_*_*.txt")
+
+    tft_transformer = TftTransformer()
+    tft_transformer.parse(args.input)
+    tft_transformer.export(args.output, lang_files, True)
+
+    #py tft.py -o D:/tft/ D:/pbe/
```
benoitryder: This comment is not needed.

PR #28, opened 2019-10-25
Update tft data script to support multiple sets

stelar7:
Had some wierd issue that str couldnt be compared with null or something, so turned off the sorting.
I _think_ everything should be correct, but i only did a quick check.

These are breaking changes, so a compatibility patch is needed to merge them

### DIFF
--- a/cdragontoolbox/tftdata.py
+++ b/cdragontoolbox/tftdata.py
@@ -122,10 +122,7 @@ class TftTransformer:
 
         if not set_collection:
             champion_list = max(character_lists.values(), key=lambda coll: len(coll.fields[0].value)).fields[0].value
-            set_characters = []
-
-            for char in champion_list:
-                set_characters.append(character_lookup[char] if char in character_lookup else None)
+            set_characters = [character_lookup.get(char) for char in champion_list]
 
             sets[1] = {
                 "name": "Base",
--- a/cdragontoolbox/tftdata.py
+++ b/cdragontoolbox/tftdata.py
@@ -1,3 +1,4 @@
+import math
 import json
 import glob
 import os
@@ -23,15 +24,16 @@ class TftTransformer:
         character_folder = os.path.join(self.input_dir, "data", "characters")
 
         map22 = BinFile(map22_file)
-        items = self.parse_items(map22)
+        character_lookup = self.parse_lookup_table(map22)
+        sets = self.parse_sets(map22, character_lookup)
         traits = self.parse_traits(map22)
         champs = self.parse_champs(map22, traits, character_folder)
+        output_sets = self.generate_correct_set_info(sets, traits, champs)
 
-        # clean up data for JSON export
-        [x.pop("internal") for x in traits]
+        items = self.parse_items(map22)
         [x.pop("internal") for x in items]
 
-        return {"items": items, "traits": traits, "champs": champs}
+        return {"sets": output_sets, "items": items}
 
     def export(self, output, langs=None):
         """Export TFT data for given languages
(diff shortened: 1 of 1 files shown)


cdragontoolbox/tftdata.py
```diff
@@ -65,17 +67,70 @@ def replace_in_data(data):
                     if key in data and data[key] in replacements:
                         data[key] = replacements[data[key]]
 
-            for data in template["champs"]:
+            for data in [template["sets"][x]["traits"] for x in template["sets"]]:
+                replace_in_data(data)
+            for data in [template["sets"][x]["champions"] for x in template["sets"]]:
```
benoitryder: - No need to create a list, the generator is sufficient
- You can iterate on values directly
```py
for data in (x["champions"] for x in template["sets"].values()):
```
(Same for the two `for` loops)

cdragontoolbox/tftdata.py
```diff
@@ -65,17 +67,70 @@ def replace_in_data(data):
...
+
+        set_map = {}
+        for key, value in sets.items():
+            current_set = {"name": value["name"]}
+
+            set_champs = []
+            for char_name in value["internal"]["characters"]:
+                for real_char in champs:
+                    if char_name == real_char["internal"]["name"]:
+                        set_champs.append(real_char)
```
benoitryder: As for previous PR, you don't need those `internal` entries. `champs` is only used for its "internal names" and `sets` is only used for its "internal characters", so no need to mix them with other values.

Furthermore, this will ease computations made in this function.

cdragontoolbox/tftdata.py
```diff
@@ -65,17 +67,70 @@ def replace_in_data(data):
...
+
+            set_traits = set([y for x in set_champs for y in x["traits"]])
```
benoitryder: No need to convert `set()` arg to a list.
And you don't even need `set()`: `{y for x in set_champs for y in x["traits"]}`

cdragontoolbox/tftdata.py
```diff
@@ -65,17 +67,70 @@ def replace_in_data(data):
...
+        for item in set_collection:
+            char_list = item.getv("characterLists")[0]
+            set_info = item[0xD2538E5A].value
+            set_no = set_info["SetNumber"]["mValue"].value
+            set_name = set_info["SetName"]["mValue"].value
+            set_characters = []
+
+            character_lists = [x for x in map22.entries if x.type == "MapCharacterList"]
+            for char_list_iter in character_lists:
+                if not char_list_iter.path == char_list:
```
benoitryder: `x != y` is preferred over `not x == y`

cdragontoolbox/tftdata.py
```diff
@@ -65,17 +67,70 @@ def replace_in_data(data):
...
+        sets = {}
+
+        for item in set_collection:
+            char_list = item.getv("characterLists")[0]
+            set_info = item[0xD2538E5A].value
+            set_no = set_info["SetNumber"]["mValue"].value
+            set_name = set_info["SetName"]["mValue"].value
+            set_characters = []
+
+            character_lists = [x for x in map22.entries if x.type == "MapCharacterList"]
```
benoitryder: No need to recompute this in every iteration. It will not change.

And I think you could create a dict indexed by `x.path` directly, to speed-up looking (the `for` loop below).
Unless there are duplicate `x.path` (I don't know if it's possible).

cdragontoolbox/tftdata.py
```diff
@@ -86,9 +141,11 @@ def parse_items(self, map22):
             if "Template" in name or name == "TFT_Item_Null":
                 continue
 
-            effects = []
+            effects = {}
             for effect in item.getv("effectAmounts", []):
-                effects.append({"name": effect.getv("name"), "value": effect.getv("value")})
+                name = effect.getv("name").s if effect.__contains__("name") else "null"
```
benoitryder: Don't call `__contains__` directly, use `"name" in effect`.

cdragontoolbox/tftdata.py
```diff
@@ -86,9 +141,11 @@ def parse_items(self, map22):
...
+                value = effect.getv("value", "null")
+                effects[name] = value
```
benoitryder: No need for the intermediate `value` variable.

cdragontoolbox/tftdata.py
```diff
@@ -108,29 +165,44 @@ def parse_items(self, map22):
...
+    def find_closest_path(self, parent, name):
+        test = name
+        while test:
+            path = os.path.join(parent, test, test + ".bin")
+            if os.path.exists(path):
+                return path
+
+            test = test[:-1]
+
+        return ""
```
benoitryder: You should return `None` for "nothing", or even raise an error since the case does not seem to be handled anyway.

cdragontoolbox/tftdata.py
```diff
@@ -108,29 +165,44 @@ def parse_items(self, map22):
...
-
         for champ in champ_collection:
             name = champ.getv("mName")
 
-            if not name.startswith("TFT_") or name == "TFT_Template":
+            if name in ["TFT_Template", "Sold", "SellAction", "GetXPAction", "RerollAction", "LockAction"]:
                 continue
 
-            char_bin = BinFile(os.path.join(character_folder, name[4:], name[4:] + ".bin").lower())
+            real_name = name[name.rfind("_") + 1 :]
```
benoitryder: `real_name = name.rsplit("_", 1)[-1]|

cdragontoolbox/tftdata.py
```diff
@@ -180,15 +255,20 @@ def parse_traits(self, map22):
 
             effects = []
             for trait_set in trait.getv("mTraitSets"):
-                variables = []
+                variables = {}
+                increment = 0
                 for effect in trait_set.getv("effectAmounts", []):
-                    variables.append({"name": effect.getv("name"), "value": effect.getv("value")})
+                    name = effect.getv("name").s if effect.__contains__("name") else "unknown" + str(increment)
```
benoitryder: - use `in` instead of `__contains__`
- Use f-string: `f"unknown{increment}"`

cdragontoolbox/tftdata.py
```diff
@@ -180,15 +255,20 @@ def parse_traits(self, map22):
...
+                    name = "unknown" + str(increment) if name == "null" else "unknown" + str(increment)
```
benoitryder: The ternary returns the same value in both cases.

Moreover, you should probably use `None` instead of `"null"` as "nothing" value.

cdragontoolbox/tftdata.py
```diff
@@ -180,15 +255,20 @@ def parse_traits(self, map22):
...
+                    value = effect.getv("value", "null")
+                    variables[name] = value
+                    increment = increment + 1
```
benoitryder: You can use `increment += 1`

cdragontoolbox/tftdata.py
stelar7: this should use the hash path instead of unknown, i just couldnt be bothered to find what variable it was hidden in

cdragontoolbox/tftdata.py
stelar7: I'm not sure what change you would propose here?

benoitryder: Instead of having `sets = {key: {"internal": internal_data, "other": ...}}` you can have `sets = {key: internal_data}` directly.
You should be able to do something similar for the other `"internal"` (the one in`traits`).

cdragontoolbox/tftdata.py
benoitryder: I don't see the link with my comments (especially the one about `__contains__`).

cdragontoolbox/tftdata.py
stelar7: In both of the cases, the parent object pass along other data. so i dont see that as possible?

benoitryder: The non-internal data does not seem to be used, so you don't need to merge internal and non-internal data.

stelar7: in both cases the `internal` data is used as a lookup object to merge the full data into the set, so it is used.

benoitryder: But the full data is not from `sets` is not merged (only `"name"` is used, I missed this one at first).
The data that is merged comes from `champs`, for which you also use an `"internal"` (which would be a more difficult to remove, I agree).

With the current design, code flow (more accurately, the data flow) is difficult to follow. It abuses mutability of data sets that should be immutable. The method return a built value but also modifies its parameters as a side effect.

Design-wise, I don't see a good reason to put `sets[key]["internal"]` into `sets[key]`; both are of different nature and should not be mixed.
You can either store pairs of internal/non-internal data (something like `sets[key] = (not_internal, internal)`)`, or use separate variables with the same indexing. And that may allow you to use a different indexing for the internal data, in order to avoid some of the "real char name" lookups.

stelar7: imo its the same if you store 
`sets[x][internal] = a 
sets[x][real] = b` 
vs 
`sets[x] = (internal, real)`

benoitryder: I agree, but in your case you use neither.
You have `sets[x] = b` and `b["internal"] = a`

And there is also the fact that you update the parameters, which IMHO is the biggest issue. Having the `"internal"` field wouldn't be an issue if it was just considered as part of the data instead of being added temporarily before being hacked away.

benoitryder:
I made various changes to your proposal branch:
- requested changes (including removal of the `internal` fields)
- additional code clean-up (including to previous code parts)
- merge to master (the PR branch was conflicting)
- various fixes to the output data (including restoring key sorting in the JSON output)

Changes have been pushed to master.

PR #29, opened 2019-11-17
Speedup xxhash calls

Morilli:
My testing showed a speed increase of about 33% (going from 2 minutes to 1.30).

### DIFF
--- a/cdragontoolbox/hashes.py
+++ b/cdragontoolbox/hashes.py
@@ -10,7 +10,7 @@ import struct
 import logging
 from contextlib import contextmanager
 from typing import Dict
-from xxhash import xxh64_intdigest as intdigest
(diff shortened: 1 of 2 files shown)


cdragontoolbox/hashes.py
```diff
@@ -10,7 +10,7 @@
 import logging
 from contextlib import contextmanager
 from typing import Dict
-from xxhash import xxh64
+from xxhash import xxh64_intdigest as intdigest
```
benoitryder: It would be better to keep the full `xxh64_intdigest` name, for clarity and because each hash method usually provides its own `intdigest` and it's better to avoid name conflicts.

Morilli: 👍

PR #33, opened 2020-02-28
add hashes

moonshadow565:
generated by dumping strings from executables and demangling any symbol names found
used a simple regex to extract words: ``(?:[a-zA-Z0-9]+_?)+[a-zA-Z0-9]+``
unfiltered .txt containing source strings can be found here: https://mega.nz/#!u90xQaDI!u7tbR64jZNqKuhl1NA_Jzkga1KbkC3yGAlnuIYCnWN8

### DIFF
--- a/cdragontoolbox/hashes.binfields.txt
+++ b/cdragontoolbox/hashes.binfields.txt
@@ -30,6 +30,7 @@ a2621347 BarrackMinion
 d488fdc9 BasedOnMap
 4948cbcb BattleFury
 a1bef65d BaublesSpell
+091fa8ec BitIndex
 a2734825 Blinded
 c967b34f Bloodwell
 25834918 BlueTeamElementSelectedColor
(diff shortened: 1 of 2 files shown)


benoitryder:
The diff removes 3 hashes from `hashes.binfields.txt`. I assume it's not intended?

moonshadow565:
Its removed because it seems a duplicate sliped through before :)
https://github.com/CommunityDragon/CDTB/blob/master/cdragontoolbox/hashes.binfields.txt#L50
https://github.com/CommunityDragon/CDTB/blob/master/cdragontoolbox/hashes.binfields.txt#L639

moonshadow565:
This commit: https://github.com/CommunityDragon/CDTB/commit/d580e7e69e4c9c4b23ff71062c18a48b163afe0c#diff-c0b3f92d9d53dbca557e1f9e841eb15d

benoitryder:
Indeed. Good catch.

PR #36, opened 2020-05-28
Fix check_basename_prefixes for when a path doesn't have a directory

Morilli:


cdragontoolbox/hashes.py
```diff
@@ -540,8 +540,8 @@ def check_basename_prefixes(self, prefixes=None):
         if prefixes is None:
             prefixes = ['2x_', '2x_sd_', '4x_', '4x_sd_', 'sd_']
         for p in self.known.values():
-            path, basename = p.rsplit('/', 1)
-            values.update(f"{path}/{prefix}{basename}" for prefix in prefixes)
+            path, basename = os.path.split(p)
+            values.update(f"{f'{path}/' if path else ''}{prefix}{basename}" for prefix in prefixes)
```
benoitryder, line 544: To optimize a bit you can factorize the ternary operator (and more importantly, appending the `/`):
```py
path, basename = os.path.split(p)
if path:
    path += '/'
values.update(f"{path}{prefix}{basename}" for prefix in prefixes)
```

Morilli: right, forgot about the empty default

PR #37, opened 2020-06-10
Fix attack range extraction in the TFT data extraction.

trebonius0:
In some cases, starting season 3.5, the attackRange field can be empty in the extracted files


cdragontoolbox/tftdata.py
```diff
@@ -220,7 +220,7 @@ def parse_champs(self, map22, traits, character_folder):
                     "critMultiplier": record.getv("critDamageMultiplier"),
                     "critChance": record.getv("baseCritChance"),
                     "attackSpeed": record.getv("attackSpeed"),
-                    "range": record["attackRange"].value // 180,
+                    "range": record.getv("attackRange", 0) // 180,
```
benoitryder, line 223: Using `getv("attackRange")` (returning `None` if there is no value) would be better as it would be consistent with other fields.

PR #42, opened 2020-10-27
Add support for .metal shader file guessing

Morilli:
Also modified the logic to support .ps_3_0 and .vs_3_0 as well

### DIFF
--- a/cdragontoolbox/hashes.py
+++ b/cdragontoolbox/hashes.py
@@ -722,7 +722,8 @@ class GameHashGuesser(HashGuesser):
     def guess_shader_variants(self):
         """Guess different extension variants for shader files, e.g. ".glsl_100" """
 
-        shader_paths = [path for path in self.known.values() if os.path.splitext(path)[-1] in [f".{variant}s_{n}_0" for variant in "pv" for n in "23"]]
+        shader_extensions = [f".{variant}s_{n}_0" for variant in "pv" for n in "23"]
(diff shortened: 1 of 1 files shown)


stelar7:
im guessing its refering to the shader version, and in that case it prolly goes beyond 3, so a generic check for all versions from 0.0 to 10.10 would be better imo

Morilli:
Hm well the game doesn't use any that are beyond version 3, but the official (i think?) documentation lists up to version 5: https://developer.download.nvidia.com/cg/index_profiles.html
I guess we could bump the number to 5, but I don't really see the point currently as there are zero hits for that, and other parts of the code likely should be adjusted as well in that case.

cdragontoolbox/hashes.py
```diff
@@ -722,15 +722,16 @@ def guess_characters_files(self, chars=None):
     def guess_shader_variants(self):
         """Guess different extension variants for shader files, e.g. ".glsl_100" """
 
-        shader_paths = [path for path in self.known.values() if path.endswith(".ps_2_0") or path.endswith(".vs_2_0")]
-
+        shader_paths = [path for path in self.known.values() if os.path.splitext(path)[-1] in [f".{variant}s_{n}_0" for variant in "pv" for n in "23"]]
```
benoitryder, line 725: You should use a temporary for `[f".{variant}s_{n}_0" for variant in "pv" for n in "23"]`, for readability and to avoid recomputing the list for each path.

Morilli: Good now?

PR #45, opened 2021-03-26
Update hashes

Jochem-W:
I used the following PowerShell script to recursively guess the hashes for each wad.client file:
```powershell
Get-ChildItem -Path 'C:\Riot Games\League of Legends\' -Filter *.wad.client -Recurse | ForEach-Object {
    python.exe -m cdragontoolbox hashes-guess $_.FullName
}
```


benoitryder:
I just updated hashes and pushed them. The update includes all hashes from this PR (among other ones).

PR #50, opened 2021-04-14
Add support for the patch section in PTCH-bin files

Morilli:
yeah well should work i think

credit to moonshadow for format and stuff

### DIFF
--- a/cdragontoolbox/binfile.py
+++ b/cdragontoolbox/binfile.py
@@ -371,7 +371,10 @@ class BinFile:
             self.patch_entries = None
 
     def to_serializable(self):
-        return (({entry.path.to_serializable(): entry.to_serializable() for entry in self.entries},) + (({entry.path.to_serializable(): entry.to_serializable() for entry in self.patch_entries},) if self.patch_entries else ()))
+        serialized = {entry.path.to_serializable(): entry.to_serializable() for entry in self.entries}
+        if self.patch_entries is not None:
+            serialized["__patches"] = {entry.path.to_serializable(): entry.to_serializable() for entry in self.patch_entries}
+        return serialized
 
 
 class BinReader:
--- a/cdragontoolbox/binfile.py
+++ b/cdragontoolbox/binfile.py
@@ -332,14 +332,17 @@ class BinNestedField(BinField):
         return (self.name.to_serializable(), self.value.to_serializable())
 
 
-class BinPtchEntry(BinObjectWithFields):
-    def __init__(self, hpath, fields):
+class BinPtchEntry:
+    def __init__(self, hpath, value):
         self.path = BinEntryPath(hpath)
-        super().__init__(fields)
+        self.value = value
 
     def __repr__(self):
-        sfields = _repr_indent_list(self.fields)
+        sfields = _repr_indent_list(self.value.fields)
         return f"<BinPtchEntry {self.path!r} {sfields}>"
+
+    def to_serializable(self):
+        return self.value.to_serializable()
 
 class BinEntry(BinObjectWithFieldsAndType):
     def __init__(self, hpath, htype, fields):
@@ -426,12 +429,9 @@ class BinReader:
             parts = objectpath.split('.')
             binvalue = self.read_field_basic(compute_binhash(parts[-1]), btype)
             if hpath not in patch_entries:
-                if len(parts) > 1:
-                    patch_entries[hpath] = BinPtchEntry(hpath, [BinNestedField(compute_binhash(parts[0]), BinNested([]))])
-                else:
-                    patch_entries[hpath] = BinPtchEntry(hpath, [])
+                patch_entries[hpath] = BinPtchEntry(hpath, BinNested([]))
 
-            current_nesting = patch_entries[hpath]
+            current_nesting = patch_entries[hpath].value
             for i in range(0, len(parts)-1):
                 if parts[i] not in current_nesting:
                     current_nesting[parts[i]] = BinNestedField(compute_binhash(parts[i]), BinNested([]))
--- a/cdragontoolbox/binfile.py
+++ b/cdragontoolbox/binfile.py
@@ -432,10 +432,10 @@ class BinReader:
                 patch_entries[hpath] = BinPtchEntry(hpath, BinNested([]))
 
             current_nesting = patch_entries[hpath].value
-            for i in range(0, len(parts)-1):
-                if parts[i] not in current_nesting:
-                    current_nesting[parts[i]] = BinNestedField(compute_binhash(parts[i]), BinNested([]))
-                current_nesting = current_nesting.getv(parts[i])
+            for part in parts[:-1]:
+                if part not in current_nesting:
+                    current_nesting[part] = BinNestedField(compute_binhash(part), BinNested([]))
+                current_nesting = current_nesting.getv(part)
 
             current_nesting[parts[-1]] = binvalue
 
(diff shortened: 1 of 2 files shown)


cdragontoolbox/binfile.py
```diff
@@ -361,6 +402,33 @@ def read_binfile_entry(self, htype):
...
+            if hpath not in patch_entries:
+                if len(parts) > 1:
+                    patch_entries[hpath] = BinPtchEntry(hpath, [BinNestedField(compute_binhash(parts[0]), BinNested([]))])
+                else:
+                    patch_entries[hpath] = BinPtchEntry(hpath, [])
+
+            current_nesting = patch_entries[hpath]
+            for i in range(0, len(parts)-1):
+                if parts[i] not in current_nesting:
+                    current_nesting.fields.append(BinNestedField(compute_binhash(parts[i]), BinNested([])))
```
benoitryder, line 425: You could define `__setitem__` on the base class (along with the existing `__getitem__`) for nicer way to set fields.

cdragontoolbox/binfile.py
```diff
@@ -294,6 +294,44 @@ def __repr__(self):
     def to_serializable(self):
         return (self.name.to_serializable(), {_to_serializable(k): _to_serializable(v) for k,v in self.value.items()})
 
+class BinNested(BinObjectWithFields):
+    """Nested binary value"""
+
+    def __init__(self, fields):
+        super().__init__(0, fields)
```
benoitryder, line 301: To avoid this useless/hacky `0` you should define a new base class but without `htype`.
Typically:
- remove `type` from `BinObjectWithFields` and adjust its `to_serializable()` accordingly
- define a new `BinObjectWithFieldsAndType` derived from `BinObjectWithFields` and adding the `type` attribute
- change previous uses of `BinObjectWithFields` to now inherit from `BinObjectWithFieldsAndType`
- remove custom constructor from `BinNested` and `BinNestedField` (not needed anymore)

benoitryder (changes requested):

cdragontoolbox/binfile.py
```diff
@@ -319,10 +357,13 @@ def __init__(self, f, btype_version=None):
         reader = BinReader(f, btype_version=btype_version)
         self.version, self.linked_files, entry_types = reader.read_binfile_header()
         self.entries = [reader.read_binfile_entry(htype) for htype in entry_types]
+        if self.is_patch and self.version >= 3:
```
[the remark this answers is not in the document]
Morilli: Both exists, files that only have patch entries and files that have both patch entries and "normal" entries.

Example for both entry types: `ux/renderui/overrides/default/apheliosui.bin`
example for only ptch entries: `gameplay/voicechat.uitft.bin`
(both are in UI.wad.client)

cdragontoolbox/binfile.py
```diff
@@ -361,6 +402,33 @@ def read_binfile_entry(self, htype):
...
 
+    def read_patch_section(self):
+        """Reads the entire PTCH section"""
+
+        count = self.read_u32()
+        patch_entries = {}
+        for _ in range(count):
+            hpath = self.read_fmt('<2L')[0]
+            btype = self.parse_bintype(self.read_u8())
+            string = self.read_string()
```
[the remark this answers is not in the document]
Morilli: Not sure how to describe it, it's like a nested class name, e.g. `mAtlas.mTextureSourceResolutionHeight`.
Probably something like `objectpath`?

It would also be possible to not interpret a `.` as hierarchy change, but I believe that makes sense, especially seeing as how other structs with the name `mAtlas` for example exist.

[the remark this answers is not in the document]
benoitryder: `objectpath` seems nice. 👍

cdragontoolbox/binfile.py
```diff
@@ -319,10 +357,13 @@ def __init__(self, f, btype_version=None):
...
         self.entries = [reader.read_binfile_entry(htype) for htype in entry_types]
+        if self.is_patch and self.version >= 3:
+            self.patch_entries = reader.read_patch_section()
+        else:
+            self.patch_entries = None
 
     def to_serializable(self):
-        return {entry.path.to_serializable(): entry.to_serializable() for entry in self.entries}
-
+        return (({entry.path.to_serializable(): entry.to_serializable() for entry in self.entries},) + (({entry.path.to_serializable(): entry.to_serializable() for entry in self.patch_entries}) if self.patch_entries else ()))
```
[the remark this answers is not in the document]
Morilli: How about 1db86f27a809ec9a6ada3e35a181328b3ea4ce90?

cdragontoolbox/binfile.py
```diff
@@ -361,6 +414,33 @@ def read_binfile_entry(self, htype):
...
+            hpath = self.read_fmt('<2L')[0]
+            btype = self.parse_bintype(self.read_u8())
+            objectpath = self.read_string()
+            parts = objectpath.split('.')
+            binvalue = self.read_field_basic(compute_binhash(parts[-1]), btype)
+            if hpath not in patch_entries:
+                if len(parts) > 1:
+                    patch_entries[hpath] = BinPtchEntry(hpath, [BinNestedField(compute_binhash(parts[0]), BinNested([]))])
+                else:
+                    patch_entries[hpath] = BinPtchEntry(hpath, [])
```
benoitryder, line 432: The special case of the initialization can be avoided by using a `BinNested` as the top level (directly in the `BinPtchEntry`).
`BinPtchEntry` would be adjusted to simply contain two fields: `hpath` (same as current), and `value` (which would be a `BinNested`).
It should make code cleaner since the whole patch/nesting stuff will be only solely by `BinNested` and its friend `BinNestedField` (no need for special processing in `BinPtchEntry`).
Note that the same could probably be made for `BinEntry`, if it stored a `BinStruct` directly instead of a list of fields; it's outside of the scope of this PR though so don't worry about it.

For patch entries this would look like this.
```py
class BinPtchEntry:
    def __init__(self, hpath, value):
        self.path = BinEntryPath(hpath)
        self.value = value  # BinNested
    ...
    # Note: to_serializable() needs adjustements too


def read_patch_section(self):
    ...

    if hpath not in patch_entries:
        patch_entries[hpath] = BinPtchEntry(hpath, BinNested([]))
    current_nesting = patch_entries[hpath].value

    for part in range(0, len(parts)-1):
        if part not in current_nesting:
            current_nesting[part] = BinNestedField(compute_binhash(part), BinNested([]))
        current_nesting = current_nesting[part]

    current_nesting[parts[-1]] = binvalue
```

benoitryder (changes requested):

cdragontoolbox/binfile.py
Morilli: I don't understand.

cdragontoolbox/binfile.py
```diff
@@ -361,6 +417,30 @@ def read_binfile_entry(self, htype):
...
+            parts = objectpath.split('.')
+            binvalue = self.read_field_basic(compute_binhash(parts[-1]), btype)
+            if hpath not in patch_entries:
+                patch_entries[hpath] = BinPtchEntry(hpath, BinNested([]))
+
+            current_nesting = patch_entries[hpath].value
+            for i in range(0, len(parts)-1):
+                if parts[i] not in current_nesting:
+                    current_nesting[parts[i]] = BinNestedField(compute_binhash(parts[i]), BinNested([]))
+                current_nesting = current_nesting.getv(parts[i])
```
benoitryder, line 438: You can iterate directly on parts using `for part in parts[:-1]:` then using `part` instead of `parts[i]`.
(I included that in my previous snippet but forgot to change the `for`. I Should have explicit that additional change. :( )

PR #52, opened 2021-04-18
add bin version argument to the bin-dump command and fix early bin versions

Morilli:
might work

### DIFF
--- a/cdragontoolbox/__main__.py
+++ b/cdragontoolbox/__main__.py
@@ -302,7 +302,7 @@ def command_bin_dump(parser, args):
         parser.error(f"BIN file not found: {args.bin}")
 
     with open(args.bin, 'rb') as f:
-        binfile = BinFile(f)
+        binfile = BinFile(f, btype_version=args.version)
     if args.json:
         json.dump(binfile.to_serializable(), sys.stdout)
     else:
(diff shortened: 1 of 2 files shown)


cdragontoolbox/__main__.py
```diff
@@ -441,6 +441,8 @@ def create_parser():
                                       help="dump a BIN file as a text tree")
     subparser.add_argument('-j', '--json', action='store_true',
                            help="extract to JSON")
+    subparser.add_argument('-V', '--version', default=1008, type=int,
+                           help="patch version this BIN file belongs to (default: %(default)s)")
```
benoitryder, line 445: - The full option name could be `--patch-version` to be more explicit about what it should contain.
- It would be nice to accept `10.08` instead of `1008`. Internally we use an integer to ease comparison, but if the CLI argument is intended to receive a patch version (as it is documented), you should accept the doted version number and convert it afterwards.

cdragontoolbox/binfile.py
```diff
@@ -482,7 +482,7 @@ def parse_bintype(self, v):
                 v = 0x80
             elif v >= 19:
                 v = 0x80 + v - 18
-        elif self.btype_version < 1008:
+        if self.btype_version < 1008:
```
benoitryder, line 485: Why this change?

benoitryder (changes requested):

cdragontoolbox/binfile.py
Morilli: Because early bins failed to parse without it.

cdragontoolbox/__main__.py
Morilli: 4fe04daf6aa4c0b9f8f1bdb1777b2d17e2f8b8b4 fine? that will allow both numbers like right now and normal version numbers with a single `.`, e.g. `9.2`.

PR #55, opened 2021-07-31
Add rst -> json conversion in export, add rst hashfile

Morilli:


cdragontoolbox/export.py
```diff
@@ -659,3 +662,31 @@ def convert(self, fin, output, path):
...
+        bit_mask = (1 << rstfile.hash_bits) - 1
+        hashes = {hash & bit_mask: value for hash, value in self.hashes.items()}
+        rst_json = {}
+        if rstfile.font_config:
+            rst_json["__fontconfig"] = rstfile.font_config
+        for key, value in rstfile.entries.items():
+            if key in hashes:
+                key = hashes[key]
+            rst_json[key] = value
+        rst_json["__version"] = rstfile.version
```
benoitryder, line 689: Since this is a newly defined JSON format we may avoid the `__special` entries and just put the actual map into its dedicated entry.
Somethink like this
```
{
  "entries": {"some_key": "Some translation", ...},
  ...
}
```

Also, I wonder if we should include `fontconfig` and `version`. `fontconfig` is some obscure blob, not really parsed. And `version` is only relevant for parsing AFAIK, so we don't really need it at this point.

We could decide to add them in the future if they prove to be useful; but it could be wiser to not commit to much on the format for now.

cdragontoolbox/export.py
```diff
@@ -659,3 +662,31 @@ def convert(self, fin, output, path):
...
+
+        with write_file_or_remove(output_path) as fout:
+            fout.write(json.dumps(rst_json, ensure_ascii=False).encode('utf-8'))
```
benoitryder, line 692: To avoid superfluous conversions, open the file as binary:
```py
        with write_file_or_remove(output_path, True) as fout:
            fout.write(json.dumps(rst_json, ensure_ascii=False))
```

benoitryder (changes requested):

cdragontoolbox/export.py
```diff
@@ -422,6 +424,7 @@ def _create_exporter(self, patch):
             ImageConverter(('.dds', '.tga')),
             BinConverter(re.compile(r'game/.*\.bin$'), btype_version),
             SknConverter(),
+            RstConverter(re.compile(r'game/data/menu/.*\.(txt|stringtable)$'))
```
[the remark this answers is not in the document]
Morilli: `data/menu/bootstrap_en_us.stringtable` for example, in `Bootstrap.wad.client`.
I believe they added those after version 2 of RST, aka the binary format.
I guess `.stringtable` is the "correct" extension, but because they kept the original (.txt) ones and didn't rename them, we must use this somewhat ugly heuristic to catch the `data/menu/fontconfig_en_us.txt` and similar as well.

cdragontoolbox/export.py
Morilli: I specifically added `version` because of the hash. For version 4, we have 39 bits, and for version 3 and lower 40 bits.
It's impossible to know whether a hash has the 40s bit set to 0 or is a 39-bit length hash just by looking at the number, which is what I'm mainly targeting with this thing.
Including fontconfig is also mainly for not losing information that was originally in the binary file.

cdragontoolbox/export.py
Morilli: Actually looking at it, it seems binary mode is the default and setting it to `False` could save me the `.encode('utf-8')`. Although that is working for me, I'm not sure it's save in regards to locales and stuff because I have no idea what happens when a file gets written to in non-binary mode...

benoitryder: Oh sorry, I forgot it was the default and read your code backwards. :(

Using `False` should be a bit better as it will encode on-the-fly in the `write()` call instead of encoding the whole string at once.
But more importantly, it avoids juggling with encoding in this part of the code, which is cleaner.

cdragontoolbox/export.py
benoitryder: Ok for `version`. For `fontconfig` we can keep it as it is... and maybe rework it one day. But since I doubt anyone will be using it anyway, that would be safer to just not put it. Anyone is interested can parse the original file.

Morilli: I guess I'll just the fontconfig then. It only exists for old files and seems to contain outdated information (always `[FontConfig "English"]` for all languages) anyway.

cdragontoolbox/export.py
```diff
@@ -659,3 +661,28 @@ def convert(self, fin, output, path):
...
+    def convert(self, fin, output, path):
+        output_path = os.path.join(output, path + '.json')
+        rstfile = RstFile(fin)
+        hashes = {key_to_rsthash(hash, rstfile.hash_bits): value for hash, value in self.hashes.items()}
+        rst_json = {"entries": {}}
+        for key, value in rstfile.entries.items():
+            if key in hashes:
+                key = hashes[key]
+            rst_json["entries"][key] = value
+        rst_json["version"] = rstfile.version
```
benoitryder, line 685: You could initialize this value right when initializing `rst_json`.

cdragontoolbox/export.py
```diff
@@ -659,3 +661,28 @@ def convert(self, fin, output, path):
...
+
+    def convert(self, fin, output, path):
+        output_path = os.path.join(output, path + '.json')
+        rstfile = RstFile(fin)
+        hashes = {key_to_rsthash(hash, rstfile.hash_bits): value for hash, value in self.hashes.items()}
+        rst_json = {"entries": {}}
+        for key, value in rstfile.entries.items():
+            if key in hashes:
+                key = hashes[key]
+            rst_json["entries"][key] = value
```
benoitryder, line 684: I overlooked this at first, but you should use the same format as usual for unresolved hashes:
```py
if key in hashes:
    key = hashes[key]
else:
    key = f"{{{key:10x}}}"
```
Note: I used `10x` assuming 10 hex digit max.

benoitryder (changes requested):

PR #58, opened 2021-11-09
Fix shader guessing consistently failing to find all shader files

Morilli:
- also adds the new `.dx9sm3` extension. Might wanna consider combining all the hardcoded paths into a for() expression at this point..

### DIFF
--- a/cdragontoolbox/hashes.py
+++ b/cdragontoolbox/hashes.py
@@ -722,19 +722,12 @@ class GameHashGuesser(HashGuesser):
     def guess_shader_variants(self):
         """Guess different extension variants for shader files, e.g. ".glsl_100" """
 
-        shader_extensions = [f".{variant}s_{n}_0" for variant in "pv" for n in "23"]
-        shader_paths = [path for path in self.known.values() if any(extension in path for extension in shader_extensions)]
-        for path in shader_paths:
-            self.check(f"{path}.dx9")
-            self.check(f"{path}.dx9sm3")
-            self.check(f"{path}.dx11")
-            self.check(f"{path}.glsl")
-            self.check(f"{path}.metal")
-            self.check_iter(f"{path}.dx9_{n}" for n in range(0, 100000, 100))
-            self.check_iter(f"{path}.dx9sm3_{n}" for n in range(0, 100000, 100))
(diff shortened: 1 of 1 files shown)


Morilli:
Which "shared extension detection" do you mean here? This part?
```py
shader_paths = [path for path in self.known.values() if any(extension in path for extension in shader_extensions)]
```

benoitryder:
Yes. `shader_extensions` and the `any(...)` could be replaced by using a regexp.

Morilli:
Okay, have refactored a fair bit, and it works much better now imo.

cdragontoolbox/hashes.py
```diff
@@ -722,17 +722,12 @@ def guess_characters_files(self, chars=None):
...
-            self.check(f"{path}.glsl")
-            self.check(f"{path}.metal")
-            self.check_iter(f"{path}.dx9_{n}" for n in range(0, 100000, 100))
-            self.check_iter(f"{path}.dx11_{n}" for n in range(0, 100000, 100))
-            self.check_iter(f"{path}.glsl_{n}" for n in range(0, 100000, 100))
-            self.check_iter(f"{path}.metal_{n}" for n in range(0, 100000, 100))
+        shader_pattern = re.compile(r'.*\.[pv]s_[23]_0')
+        shader_paths = {match.group(0) for path in self.known.values() if (match := shader_pattern.match(path)) is not None}
+        shader_extensions = ["dx9", "dx9sm3", "dx11", "glsl", "metal"]
+        for path in sorted(shader_paths):
```
benoitryder, line 728: I don't think sorting is required. Does it improve the output somehow?

Morilli: Sorting is not necessarily required here, but it is done in similar locations (mainly the progress iterator calls), it does sort the output for found hashes which feels more natural to me, and because I have to deduplicate by using a set earlier I have to call at least `list()` anyway, so I just made it a `sort()` instead. Can remove if wanted.

benoitryder: The list is short enough (at least for now), so that's fine.

FYI you don't need to use `list()` here: it's perfectly fine to iterate directly on the set. ;)

PR #61, opened 2022-02-22
Add apiName to items

bangingheads:
With the Riot Games API now supporting augments, the TFT data is missing the key for augments as well as they are buried in the items.

This PR separates augments from items and gives the items the key they need to match from the API.

If we believe this is too breaking of a change we can only add the key to items.

### DIFF
--- a/cdragontoolbox/tftdata.py
+++ b/cdragontoolbox/tftdata.py
@@ -48,11 +48,13 @@ class TftTransformer:
         champs = self.parse_champs(map22, traits, character_folder)
         output_sets, output_set_data = self.build_output_sets(sets, traits, champs)
         items = self.parse_items(map22)
+        augments = self.parse_augments(map22)
 
         return {
             "sets": output_sets,
             "setData": output_set_data,
             "items": items,
+            "augments": augments,
         }
 
     def export(self, output, langs=None):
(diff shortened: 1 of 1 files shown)


benoitryder:
The issue with adding an "augment" collection is more about providing a stable, future-proof format.
According to the raw data, augments are not a special data type but handled as regular items.
Riot may introduce new game mechanics in the future and it would be better to not have to add new top-level entries for each one.

I checked the data and for an item to have `Augment` in its name is equivalent to have `{b72bd3bf}` in their `ItemTags`. This looks like a nice criteria. Multiple other tags are used, with only a subset of possible combinations for augments.
We could add the `ItemTags` to provided data and/or add an `augment` flag to items that have the relevant tag (to avoid the obscure hash).
I tried to reverse the hash using variations of `augment` to have a human friendly string instead of the hash but without success. (Maybe there is another technical term for augments that I could have searched for.)

Note that I'm not familiar with TFT. I don't know if other tags could be useful/relevant. Among all the tags, only `Consumable` has been reversed; I tried to guess some others but again, no success.

And `apiName` could definitely be added to all items I guess.

bangingheads:
I wasn't able to reverse anything useful either out of those hashes.

I'm just going to change this PR to add an apiName to items so that people can match the strings from the API to the cdragon files and that should be good enough for now.

PR #66, opened 2022-11-03
New --patchline argument

LeagueOfPoro:
Added support for `--patchline pbe` argument as a working alternative to `--cdn`.

### DIFF
--- a/cdragontoolbox/__main__.py
+++ b/cdragontoolbox/__main__.py
@@ -78,7 +78,7 @@ def parse_storage_args(parser, args) -> Storage:
             if conf['type'] == 'patcher':
                 conf['patchline'] = args.patchline
             else:
-                parser.error("--pbe is only supported for 'patcher' storage")
+                parser.error("--patchline is only supported for 'patcher' storage")
     return Storage.from_conf(conf)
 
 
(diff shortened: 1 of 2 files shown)


cdragontoolbox/__main__.py
```diff
@@ -74,6 +74,8 @@ def parse_storage_args(parser, args) -> Storage:
             conf['cdn'] = cdn
         elif cdn != 'default':
             parser.error("--cdn is only supported for 'rads' storage")
+        if conf['type'] == 'patcher':
+            conf['patchline'] = args.patchline
```
benoitryder, line 78: An error should be thrown if `--patchline` is set and storage type is not `patcher` (as for `--cdn`).
This is easy to do if `--patchline` defaults to `None` as suggested.

benoitryder (changes requested):
Thanks for your PR! Several people had issues with that in the past so this will definiteily be helpful. :)

benoitryder:
Keeping `--patchline` is better IMHO, because it's more generic and matches directly the storage JSON format.

And you can use the following to have same default/fallback behavior:
```py
    storage_parser.add_argument('--patchline', choices=['live', 'pbe'], default=None,
                                help="use PBE patchline")
```
```py
        if args.patchline is not None:
            if conf['type'] == 'patcher':
                conf['patchline'] = args.patchline
            else:
                parser.error("--patchline is only supported for 'patcher' storage")
```

LeagueOfPoro:
I would argue --pbe makes for a more concise command. However, matching the JSON config makes sense. 
Changed

PR #68, opened 2022-12-15
Support reading subchunked wad entries using the .subchunktoc file

Morilli:
This also fixes an exception currently occuring on PBE when parsing the `Warwick.wad.client` file (from manifest https://lol.secure.dyn.riotcdn.net/channels/public/releases/04C8D70DFF45D23B.manifest):
```py
Traceback (most recent call last):
  File "D:\GitHub\CDTB\grep_wads.py", line 34, in <module>
    wad.guess_extensions()
  File "D:\GitHub\CDTB\cdragontoolbox\wad.py", line 214, in guess_extensions
    data = wadfile.read_data(f)
  File "D:\GitHub\CDTB\cdragontoolbox\wad.py", line 97, in read_data
    return zstd_decompress(data)
pyzstd.ZstdError: Unable to decompress zstd data: Unknown frame descriptor.
```
The failing (wad-)file in particular has hash `F5B498022C64E681` (`assets/characters/warwick/skins/skin46/particles/winterblessed_warwick_colorblue_01.pie_c_12_23.tex`) and it fails because the first subchunk is uncompressed, while the second and third are.

The fact that the subchunktoc file path must be known before being able to effectively handle files is a big issue of this implementation, but I don't see how this can be done better.
I also contemplated making an own class for the subchunktoc file entries (u32 compressed size, u32 uncompressed size, u64 xxhash of data), but decided against it for now as it didn't seem necessary.

### DIFF
--- a/cdragontoolbox/wad.py
+++ b/cdragontoolbox/wad.py
@@ -32,8 +32,6 @@ _hash_to_guessed_extensions = {}
 class MalformedSubchunkException(BaseException):
     """Subchunk data is invalid or doesn't match the provided subchunktoc"""
 
-    wad_data = None
-
     def __init__(self, data):
         self.wad_data = data
 
--- a/cdragontoolbox/wad.py
+++ b/cdragontoolbox/wad.py
@@ -75,7 +75,7 @@ class WadFileHeader:
         self.path = None
         self.ext = None
 
-    def read_data(self, f, subchunkTOC=None):
+    def read_data(self, f, subchunk_toc=None):
         """Retrieve (uncompressed) data from WAD file object"""
 
         f.seek(self.offset)
@@ -94,17 +94,17 @@ class WadFileHeader:
             return zstd_decompress(data)
         elif self.type == 4:
             # Data is split into individual subchunks that may be zstd compressed
-            if subchunkTOC is not None:
-                subchunkEntries = [struct.unpack('<IIQ', subchunkTOC[16*(subchunkIndex := self.first_subchunk_index+i):16*(subchunkIndex+1)]) for i in range(self.subchunk_count)]
+            if subchunk_toc is not None:
+                subchunk_entries = [struct.unpack('<IIQ', subchunk_toc[16*(subchunk_index := self.first_subchunk_index+i):16*(subchunk_index+1)]) for i in range(self.subchunk_count)]
                 output = bytes()
                 offset = 0
-                for subchunk in subchunkEntries:
-                    subchunkData = data[offset:offset+subchunk[0]]
-                    if subchunk[0] == subchunk[1]: # assume uncompressed
-                        output += subchunkData
+                for (compressed_size, uncompressed_size, _hash) in subchunk_entries:
+                    subchunk_data = data[offset:offset+compressed_size]
+                    if compressed_size == uncompressed_size: # assume data is uncompressed
+                        output += subchunk_data
                     else:
-                        output += zstd_decompress(subchunkData)
-                    offset += subchunk[0]
+                        output += zstd_decompress(subchunk_data)
+                    offset += compressed_size
                 return output
             else:
                 try:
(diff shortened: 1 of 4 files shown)


cdragontoolbox/wad.py
```diff
@@ -93,18 +93,34 @@ def read_data(self, f):
...
             return zstd_decompress(data)
         elif self.type == 4:
-            # Data is split into individual subchunks that are zstd compressed
-            return zstd_decompress(data)
+            # Data is split into individual subchunks that may be zstd compressed
+            if subchunkTOC is not None:
+                subchunkEntries = [struct.unpack('<IIQ', subchunkTOC[16*(subchunkIndex := self.first_subchunk_index+i):16*(subchunkIndex+1)]) for i in range(self.subchunk_count)]
+                output = bytes()
+                offset = 0
+                for subchunk in subchunkEntries:
```
benoitryder, line 101: You could use `for (compressed, uncompressed, _hash) in subchunk_entries:` to at least "document" the content of the tuple (while still not bothering to create a dedicated class).

benoitryder (changes requested):

cdragontoolbox/wad.py
```diff
@@ -93,18 +93,34 @@ def read_data(self, f):
...
+                        output += subchunkData
+                    else:
+                        output += zstd_decompress(subchunkData)
+                    offset += subchunk[0]
+                return output
+            else:
+                try:
+                    return zstd_decompress(data)
+                except:
+                    return data # possibly wrong, but what else to do when the subchunkTOC isn't given and decompression fails?
```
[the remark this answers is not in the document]
Morilli: I also thought about it, and at least for our code, not throwing or at least catching any exceptions is probably a must.

Take the warwick file I mentioned as an example: In the patch they added the failing subchunked file, they also added the subchunktoc file. That means that the subchunktoc hash isn't resolved and extracting the file or even just running the `guess_extensions` function will throw, completely denying access to the wad file.

We could probably add heuristic `.subchunktoc` hash guessing code to ensure this doesn't happen, but catching it in case it does should still be done.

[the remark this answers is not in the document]
benoitryder: The using `subchunk_toc` parameter could be a good idea.
* `subchunk_toc = something`: decompress using provided subchunk data, fail on error
* `subchunk_toc = None`: assume subchunk is unknown, return the original data on error (original proposal)
* `subchunk_toc = False`: there is no suchunk data at all; fail on error

Using `False` as default would be safer, and we could use `None` internally in our code.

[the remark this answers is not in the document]
benoitryder: I should have checked your changes first. I didn't saw you used the exception approach. 👍

cdragontoolbox/wad.py
```diff
@@ -28,6 +29,13 @@ def test_jpeg_photoshop(h, f):
 # read file data, caching will reduce I/Os
 _hash_to_guessed_extensions = {}
 
+class MalformedSubchunkException(BaseException):
+    """Subchunk data is invalid or doesn't match the provided subchunktoc"""
+
+    wad_data = None
```
benoitryder, line 35: Why having this attribute on the class itself?

Morilli: I was going by your suggestion of having the invalid data stored in the exception. I don't actually know how exactly exceptions work so I don't know if there is a better way to do this.

benoitryder: I'm not an expert on the good practices for custom exceptions with data (other than messages), but it seems fine.
But you don't have to declare `wad_data` here. Using `self.wad_data` in the constructor is enough.

Morilli:
Yeah I wanted to look at this again as well. One more thing I thought of would be to try and guess possible `.subchunktoc` paths before trying to load it, perhaps in the `load_subchunk_toc` function as well.

The existing `.subchunktoc` files follow the very predictable pattern of being based on the wad's file path in the actual game files, e.g. `data/final/champions/ashe.wad.subchunktoc`, so it should be possible to consistently find it if it exists in the wad.

benoitryder:
Why not. If it becomes more complex we could also consider moving the resolve+load part in a separate function only executed when first accessing `subchunk_toc`. Not hard to do and that would avoid having to read data when the subchunks are actually never read.

benoitryder:
Commit pushed separately, with few adjustements.

PR #71, opened 2023-03-16
Add AssociatedTraits and IncompatibleTraits to TFT item and augment data

SamGuthrie:
Adds AssociatedTraits and IncompatibleTraits from map22.bin to item and augment data.

AssociatedTraits seems to be the rules that augments use for being offered, and is useful to associate an augment with the trait it gives ie. "TFT8_Augment_InterPolarisTrait" (LaserCorps Heart) gives +1 "Set8_SpaceCorps"

IncompatibleTraits (not sure why its named this but kept the name) seems to be the trait that spatula items are associated with ie "TFT8_Item_OxForceEmblemItem" (OxForce emblem) gives +1 "Set8_OxForce"

### DIFF
--- a/cdragontoolbox/tftdata.py
+++ b/cdragontoolbox/tftdata.py
@@ -178,6 +178,14 @@ class TftTransformer:
 
     def parse_items(self, map22):
         item_entries = [x for x in map22.entries if x.type == "TftItemData"]
+        trait_entries = [x for x in map22.entries if x.type == "TftTraitData"]
+
+        traits_by_hash = {}
+        for trait in trait_entries:
+            trait_apiName = trait.getv("mName")
+            if "Template" in trait_apiName:
+                continue
+            traits_by_hash[trait.path.h] = trait_apiName
 
         items = []
         items_by_hash = {}  # {item_hash: item}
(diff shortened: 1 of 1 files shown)


SamGuthrie:
Sure, should be updated in the latest commit

cdragontoolbox/tftdata.py
```diff
@@ -178,6 +178,14 @@ def parse_sets(self, map22, character_names):
 
     def parse_items(self, map22):
         item_entries = [x for x in map22.entries if x.type == "TftItemData"]
+        trait_entries = [x for x in map22.entries if x.type == "TftTraitData"]
+
+        traits_by_hash = {}
+        for trait in trait_entries:
+            trait_apiName = trait.getv("mName")
+            if "Template" in trait_apiName:
+                continue
```
benoitryder, line 187: It would be a bit safer to not filter on `Template` at all.

Since we don't list traits directly (contrary to `items`), it is not a problem if we store all of them in `traits_by_hash`.
And it is safer, as `traits_by_hash[h]` might trigger an error if a template trait is listed (unlikely, but we are never sure with Riot).

This would reduce the loop to `traits_by_hash = {trait.path.h: trait.getv("mName") for trait in trait_entries}`

benoitryder (changes requested):

SamGuthrie:
Agreed and done :)

PR #76, opened 2023-05-28
Guess extension for extensionless files and use when extracting

Morilli:
Code should be pretty self-explanatory. This should resolve potential conflicts with extensionless files and folders and also ensure those files are exported with a known good extension.

Ideally we should probably prefer the `guess_extension` method's return over a resolved path's extension, however this should be done in a way that doesn't require reading an entire wad file, potentially even multiple times, so that's for later.

### DIFF
--- a/cdragontoolbox/wad.py
+++ b/cdragontoolbox/wad.py
@@ -250,7 +250,7 @@ class Wad:
         # avoid opening the file if not needed
         unknown_ext = True
         for wadfile in self.files:
-            if not wadfile.path and not wadfile.ext:
+            if not wadfile.ext:
                 wadfile.ext = _hash_to_guessed_extensions.get(wadfile.path_hash)
                 if not wadfile.ext:
                     unknown_ext = True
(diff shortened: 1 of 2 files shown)


cdragontoolbox/wad.py
```diff
@@ -277,15 +277,20 @@ def set_unknown_paths(self, path):
...
         for wadfile in self.files:
             if wadfile.path:
                 path, filename = os.path.split(wadfile.path)
+                basename, ext = os.path.splitext(filename)
+                if wadfile.ext and not ext:
+                    # extension was guessed, but the resolved path has no extension
+                    # in this case, append the guessed extension with custom suffix
+                    wadfile.path = f"{wadfile.path}.cdtb.{wadfile.ext}"
+
                 if len(filename) < 255:
```
benoitryder, line 291: For clarity I would change this to
```py
if len(filename) >= 255:
    wadfile.path = os.path.join(path, f"{basename[:255-17-len(ext)]}.{wadfile.path_hash:016x}{ext}")
```

This makes it clearer that there are two different change being checked for and applied.

PR #80, opened 2023-06-27
Fixed tft-data json File to include all traits for a given set

Petersil1998:
I've change the code for generating the tft-data json File to include all Traits for a given Set, similar how Champions are included (instead of creating a Collection of the Champions Traits). Traits are now part of the set Tuple


benoitryder:
@Petersil1998 I just applied the fix we discussed (and tested) earlier.

Petersil1998:
Great, thanks!

PR #84, opened 2023-07-30
Improve arena/tftdata regex handling

Morilli:
This also fixes the regex to work on windows with "\\" as path separator instead of "/".

### DIFF
--- a/cdragontoolbox/arenadata.py
+++ b/cdragontoolbox/arenadata.py
@@ -40,14 +40,15 @@ class ArenaTransformer:
         """
 
         stringtable_dir = os.path.join(self.input_dir, "data/menu")
-        stringtable_pattern = "main_??_??.stringtable"
+        stringtable_glob = "main_??_??.stringtable"
         stringtable_format = "main_%s.stringtable"
+        stringtable_regex = r"main_(.._..)\.stringtable$"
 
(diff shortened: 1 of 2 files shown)


cdragontoolbox/arenadata.py
```diff
@@ -46,8 +46,8 @@ def export(self, output, langs=None):
         if langs is None:
             langs = []
             for path in glob.glob(os.path.join(stringtable_dir, stringtable_pattern)):
-                # Note: must be adjusted if format changes
-                m = re.search(r'/\w*_(.._..)\.\w*$', path)
+                # Note: may need to be adjusted if format changes
+                m = re.search(stringtable_pattern.replace('??_??', '(.._..)') + '$', path)
```
benoitryder, line 50: Building the `re` regexp by replacing the glob one can break in a lot of ways (special characters, `??_??` not being there as it is, ...).
We should probably harcode the `re` regexp along with the others at this point... Would be clearer and easier to change if needed.
Something like this, few lines above, and adjusting the uses.
```py
        stringtable_glob = "main_??_??.stringtable"
        stringtable_re = r"main_(.._..)\.stringtable$"
        stringtable_format = "main_%s.stringtable"
```

PR #90, opened 2023-12-05
Remove disabled augments from arena json

bangingheads:
Riot has decided to include disabled augments in their data with duplicated ids. 

Because of this we need to filter out disabled augments so that we have an accurate dataset.


bangingheads:
Updated.

PR #95, opened 2024-02-07
Support new stringtable path

Morilli:
This is currently pbe-only and also currently only `en_us/main.stringtable` actually exists, but PRing this early for visibility and feedback.

This could technically be merged before the current PBE paths hit live as the code now handles all 3 known path variations.

### DIFF
--- a/cdtb/arenadata.py
+++ b/cdtb/arenadata.py
@@ -1,10 +1,9 @@
 import json
-import glob
 import os
 import copy
-import re
 from .binfile import BinFile, BinHashBase
 from .rstfile import RstFile
+from .tools import stringtable_paths
 
 
 class NaiveJsonEncoder(json.JSONEncoder):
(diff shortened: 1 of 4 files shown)


cdtb/arenadata.py
```diff
@@ -40,9 +41,7 @@ def export(self, output, langs=None):
         """
 
         stringtable_dir = os.path.join(self.input_dir, "data/menu")
-        stringtable_glob = "main_??_??.stringtable"
-        stringtable_format = "main_%s.stringtable"
-        stringtable_regex = r"main_(.._..)\.stringtable$"
+        stringtable_glob, stringtable_format, stringtable_regex = stringtable_strings(stringtable_dir)
```
benoitryder, line 44: Instead of returning a tuple of those values you can return a list of `(lang, rst_file_path)`.
The method would have to accept `langs` as parameters but it's pretty simple.

cdtb/tools.py
```diff
@@ -78,3 +80,11 @@ def raw(self, length):
...
         """Unpack string prefixed by its 32-bit length"""
         return self.f.read(self.unpack('<L')[0]).decode('utf-8')
+
+def stringtable_paths(base_dir):
+    # Support all known variations of stringtable paths here:
+    # Old: 'font_config_*.txt'
+    # Newer: 'main_*.stringtable'
+    # Current: '*/main.stringtable'
+
+    return {re.search(r"(.._..)[\\/]main\.stringtable$", path).group(1): path for path in glob.glob(os.path.join(base_dir, "??_??/main.stringtable"))} or {re.search(r"main_(.._..)\.stringtable$", path).group(1): path for path in glob.glob(os.path.join(base_dir, "main_??_??.stringtable"))} or {re.search(r"fontconfig_(.._..)\.txt$", path).group(1): path for path in glob.glob(os.path.join(base_dir, "fontconfig_??_??.txt"))}
```
benoitryder, line 90: This is unreadable. :/ Spread over multiple lines or use intermediate variables.

benoitryder:
I'm fine with your PR as it is. Feel free to merge it if needed. ;)

PR #96, opened 2024-02-24
Export language-specific files to their own subfolder

Morilli:
This is mostly untested, but should work I think.

Background is that riot has renamed most language-specific files inside wads, like VO audio files, from containing their respective language's identifier (like `de_de`) to instead contain `en_us`. That means that all those files from different wads have the same name but contain different content. To avoid those conflicts while extracting language-specific files are not exported to a new folder for that language in `game/`.

### DIFF
--- a/cdtb/export.py
+++ b/cdtb/export.py
@@ -476,9 +476,13 @@ class CdragonRawPatchExporter:
 
         for path, wad in exporter.wads.items():
             if path.endswith('.wad.client'):
+                if lang_match := re.search(r"\.(.._..)\.wad\.client$", path):
+                    subdir = f"game/{lang_match.group(1).lower()}"
+                else:
+                    subdir = "game"
                 wad.files = [wf for wf in wad.files if filter_path(wf.path)]
                 for wf in wad.files:
-                    wf.path = f"game/{wf.path}"
+                    wf.path = f"{subdir}/{wf.path}"
         # remove emptied WADs
         for path, wad in list(exporter.wads.items()):
             if not wad.files:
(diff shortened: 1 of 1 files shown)


cdtb/export.py
```diff
@@ -477,8 +477,13 @@ def filter_path(path):
         for path, wad in exporter.wads.items():
             if path.endswith('.wad.client'):
                 wad.files = [wf for wf in wad.files if filter_path(wf.path)]
+                language_match = re.match(r"\.(.._..)\.wad\.client$", path)
                 for wf in wad.files:
-                    wf.path = f"game/{wf.path}"
+                    if language_match is not None:
```
benoitryder, line 482: This condition can be factorized. This is a bit cleaner and would be easier to extend if we have to introduce more special subdirs.
```py
            if path.endswith('.wad.client'):
                if lang_match := re.match(r"\.(.._..)\.wad\.client$", path):
                    subdir = f"game/{lang_match.group(1).lower()}"
                else:
                    subdir = "game"
                wad.files = [wf for wf in wad.files if filter_path(wf.path)]
                for wf in wad.files:
                    wf.path = f"{subdir}/{wf.path}"
```

benoitryder (changes requested):

cdtb/export.py
```diff
@@ -477,8 +477,13 @@ def filter_path(path):
         for path, wad in exporter.wads.items():
             if path.endswith('.wad.client'):
                 wad.files = [wf for wf in wad.files if filter_path(wf.path)]
+                language_match = re.match(r"\.(.._..)\.wad\.client$", path)
```
benoitryder, line 480: `re.mach()` → `re.search()` (or it won't work)

PR #97, opened 2024-03-19
Pull TFT character ability translation keys from character bin

bangingheads:
There are now cases like with Kayle on PBE that the ability name and description translation keys don't exist in map22, we can fallback to pulling these from the ability in the character bin.

### DIFF
--- a/cdtb/tftdata.py
+++ b/cdtb/tftdata.py
@@ -247,10 +247,16 @@ class TftTransformer:
 
             spell_name = record.getv("spellNames")[0]
             spell_name = spell_name.rsplit("/", 1)[-1].lower()
+            spell_key_name = None
+            spell_key_tooltip = None
             for entry in tft_bin.entries:
                 if entry.type == "SpellObject" and entry.getv("mScriptName").lower() == spell_name:
                     ability = entry.getv("mSpell")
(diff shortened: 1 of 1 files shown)


cdtb/tftdata.py
```diff
@@ -247,10 +247,16 @@ def parse_champs(self, map22, traits):
...
+            spell_key_name = None
+            spell_key_tooltip = None
             for entry in tft_bin.entries:
                 if entry.type == "SpellObject" and entry.getv("mScriptName").lower() == spell_name:
                     ability = entry.getv("mSpell")
                     ability_variables = [{"name": value.getv("mName"), "value": value.getv("mValues")} for value in ability.getv("mDataValues", [])]
+                    if "mClientData" in ability:
+                        loc_keys = ability.getv("mClientData").getv("mTooltipData").getv("mLocKeys")
+                        spell_key_name = loc_keys.get("keyName")
+                        spell_key_tooltip = loc_keys.get("keyTooltip")
```
benoitryder, line 259: Some `mClientData` may not have a `mTooltipData` and the script would fail.
I just pushed a method to make it easier to chain the getv.
This should now work, and protect against missing fields
```suggestion
                    if loc_keys := ability.get_path("mClientData", "mTooltipData", "mLocKeys"):
                        spell_key_name = loc_keys.get("keyName")
                        spell_key_tooltip = loc_keys.get("keyTooltip")
```

PR #99, opened 2024-07-16
Support new rst hash format (xxh3)

Morilli:
Starting with pbe patch 14.15, rst hashes now use xxh3 instead of xxh64.

I've modified the code in a similar way to the existing bin type handling, as sadly riot did not care to update the rst file version to signal the updated hashing method.

Main TODO: Decide how to handle hashes.

Currently my idea is to have a legacy hashfile containing xxh64 hashes and a new hashfile containing xxh3 hashes. That allows for backwards compatibility in parsing and converting stringtable files from before patch 14.15 using that legacy hashfile.
Current names are temporary; not entirely sure how to name them.
some ideas:
`hashes.rstlegacy.txt` - `hashes.rst.txt`
`hashes.rst.xxh64.txt` - `hashes.rst.xxh3.txt`

### DIFF
--- a/cdtb/rstfile.py
+++ b/cdtb/rstfile.py
@@ -1,4 +1,3 @@
-import os
 from xxhash import xxh3_64_intdigest, xxh64_intdigest
 from base64 import b64encode
 from .tools import BinaryParser
@@ -11,6 +10,15 @@ def get_hashfile(game_version=1415):
     else:
         return hashfile_rst_xxh64
 
+def key_to_hash(key, bits=64, rsthash_version=1415):
+    if isinstance(key, str):
+        if rsthash_version >= 1415:
+            key = xxh3_64_intdigest(key.lower())
+        else:
+            key = xxh64_intdigest(key.lower())
+    return key & ((1 << bits) - 1)
+
+
 hashfile_rst_xxh64 = HashFile(default_hash_dir / "hashes.rst.xxh64.txt", hash_size=10)
 hashfile_rst_xxh3 = HashFile(default_hash_dir / "hashes.rst.xxh3.txt", hash_size=10)
 
@@ -21,7 +29,6 @@ class RstFile:
         self.hash_bits = 40
         self.version = None
         self.rsthash_version = game_version
-        self.key_hasher = None
 
         if path_or_f is not None:
             if isinstance(path_or_f, str):
@@ -30,23 +37,16 @@ class RstFile:
             else:
                 self.parse_rst(path_or_f)
 
-    def _get_key_hasher(self):
-        if self.rsthash_version >= 1415:
-            func = xxh3_64_intdigest
-        else:
-            func = xxh64_intdigest
-        return lambda key: func(key) & ((1 << self.hash_bits) - 1)
-
     def __getitem__(self, key):
         try:
-            h = self.key_hasher(key)
+            h = key_to_hash(key, self.hash_bits, self.rsthash_version)
             return self.entries[h]
         except (TypeError, KeyError):
             raise KeyError(key)
 
     def __contains__(self, key):
         try:
-            h = self.key_hasher(key)
+            h = key_to_hash(key, self.hash_bits, self.rsthash_version)
             return h in self.entries
         except TypeError:
             return False
@@ -56,9 +56,6 @@ class RstFile:
             return self[key]
         except KeyError:
             return default
-
-    def truncate_hash(self, hash):
-        return hash & ((1 << self.hash_bits) - 1)
 
     def parse_rst(self, f):
         parser = BinaryParser(f)
@@ -80,8 +77,6 @@ class RstFile:
         else:
             raise ValueError(f"unsupported RST version: {version}")
         self.version = version
-
-        self.key_hasher = self._get_key_hasher()
 
         hash_mask = (1 << self.hash_bits) - 1
         count, = parser.unpack("<L")
--- a/cdtb/rstfile.py
+++ b/cdtb/rstfile.py
@@ -1,24 +1,34 @@
 import os
-from xxhash import xxh64_intdigest
+from xxhash import xxh3_64_intdigest, xxh64_intdigest
 from base64 import b64encode
 from .tools import BinaryParser
 from .hashes import HashFile, default_hash_dir
 
 
-def key_to_hash(key, bits=40):
+def key_to_hash(key, bits=40, rsthash_version=1415):
     if isinstance(key, str):
-        key = xxh64_intdigest(key.lower())
+        if rsthash_version >= 1415:
+            key = xxh3_64_intdigest(key.lower())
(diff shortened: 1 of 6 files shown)


benoitryder:
I like the `hashes.rst.xxh64.txt`/`hashes.rst.xxh3.txt` option. It's nice to have explicit names. Moreover, "legacy" would not work if it changes again.

cdtb/rstfile.py
```diff
@@ -1,24 +1,34 @@
 import os
-from xxhash import xxh64_intdigest
+from xxhash import xxh3_64_intdigest, xxh64_intdigest
 from base64 import b64encode
 from .tools import BinaryParser
 from .hashes import HashFile, default_hash_dir
 
 
-def key_to_hash(key, bits=40):
+def key_to_hash(key, bits=40, rsthash_version=1415):
```
benoitryder, line 8: We could use the following to limit the use of `bits`/`rsthash_version` parameters to just few places.
```py
def get_key_hasher(bits=40, rsthash_version=1415):
    if rsthash_version >= 1415:
        func = xxh3_64_intdigest
    else:
        func = xxh64_intdigest
    def key_hasher(key):
        func(key) & ((1 << bits) - 1)
    return key_hasher
```

The `key_hasher` could be retrieved as soon as possible. And the API will not have to change if we have to introduce new parameters or whatever.

For instance, for `RstFile` constructor
```py
class RstFile:
    def __init__(self, path_or_f=None, key_hasher=None):
        ...
        if key_hasher is None:
            key_hasher = get_key_hasher()
        self.key_hasher = key_hasher
```
Then use `h = self.key_hasher(key)`

cdtb/tftdata.py
```diff
@@ -345,9 +346,13 @@ def main():
     parser = argparse.ArgumentParser()
     parser.add_argument("input", help="directory with extracted bin files")
     parser.add_argument("-o", "--output", default="tft", help="output directory")
+    parser.add_argument('-V', '--patch-version', default="14.15",
```
benoitryder, line 349: We could use `99.99` or a special value (`latest` or `None`) for "last value".
This would be easier to maintain and use.

Morilli: I like having an actual patch version as the default value. It both signifies the format better than by using `99.99` (14.15 is a real patch version that exists, 99.99 isn't), and in this case carries meaning of 14.15 being a special value. The default value here is mainly for the user, not for the parser.

cdtb/rstfile.py
Morilli: hmm how would that work though? Ideally the `RstFile` class is the only place that knows what game version maps to what key function and what rst version maps to what bits.
Your suggestion assumes that the calling code is responsible for creating a key_hasher and needs to know what `bits` and `rsthash_version` to provide.

That said, we could construct something like that internally in `RstFile` to reduce `key_to_hash` invocations on every dict(-like) access.

benoitryder: > Ideally the RstFile class is the only place that knows what game version maps to what key function and what rst version maps to what bits.

The RstFile class should be generic and should not have to know about Riot shenanigans. It should be a simple container for translation strings (and their hashes/identifiers).

But more importantly: only the top-level code (updater, etc.) know about patch versions. And it's better if those concepts stay there (at high level, in the updater and its friends).
So yes, I totally assume that the calling code *must* know the "hash format" used for its file. RstFile doesn't need to know why or how hash formats are different, it just needs to know how to hash (hence the generic function parameter).
Having this intermediate type avoids to tie low-level and high-level stuff; which is usually a good idea.

We do provide a helper (`get_key_hasher()`) to help get a correct hasher. It simple to use and we can easily extend it as needed.

cdtb/tftdata.py
benoitryder: It's weird to have 14.15 since we accept any other value. Typically you could expect to provide the actual current (or next for PBE) patch value. I agree that 99.99 is ugly. But we could just accept/use `None` or `"latest"` and convert it internally to 9999 (as in `PatchVersion`).

cdtb/storage.py
```diff
@@ -88,6 +88,12 @@ def __init__(self, v: Union[str, tuple]):
                 self.s = '.'.join(str(x) for x in self.t)
             assert len(self.t) == 2, "invalid patch version format"
 
+    def __int__(self):
```
benoitryder, line 91: I'm not a fan of implementing this with `__int__`.  The returned number is not a "normal" integer that can be added to other numbers. It's some sort of special value. (We could actually use a `(x, y)` tuple btw.)

A dedicated, non-magic method would be better IMHO.

cdtb/rstfile.py
Morilli: I still don't understand how you would construct such a key_hasher from that outside before instantiating an `RstFile`. How would the calling code know the correct value for `bits`, when that information is necessarily tied to each individual `RstFile`?

cdtb/tftdata.py
Morilli: > It's weird to have 14.15 since we accept any other value.

I don't understand what's weird about this. The same is done for the `bin-dump` command. Having the default value be in the same format as what we expect an input to be is good. Users can type `--help` and get an example for what value to provide, whereas having `None` as default gives no additional information. How should the `patch-version` argument be provided? `V14.5`? That won't parse. `14.5.1`? Also won't work. I feel like this default is a perfectly natural way to convey this information.

cdtb/storage.py
benoitryder: Why not this?
```py
def version_as_int(s: str) -> int:
    if s == "main":
        return 9999
    x, y = s.split('.', 2)
    return 100 * int(x) + int(y)
```
Then use
```py
# In export code
game_version = version_as_int(self.patch.version.s)
# In arenadata.py
parsed_version = version_as_int(args.patch_version)
```
I made it simple but `version_as_int()` could be adjusted to accept more types and values (`latest`, `None`, a `PatchVersion`, etc.).

cdtb/tftdata.py
benoitryder: The `patch-version` should be set to the patch version from which the file is from. The user shouldn't have to know that `14.15` is the last version with changes to the hash format, or when the hash format changed. cdtb will change handle that for them.

In any case, we should put an example value in the help message. But this does not have to be the default value (and we wouldn't have to update it if hash format changes).

cdtb/rstfile.py
benoitryder: Hum... You're right. I forgot the bit size was read from RST header and updated after its initialization. Sorry for that.
I assumed that if there was a version, Riot would have incremented it when changing the hash format... but it's Riot. :(

IMHO defining a separate `get_key_hasher()` would be a good idea anyway.
But as you said, it cannot be computed until final bit size value is known.

PR #100, opened 2024-07-23
Support new TFT stringtable path format

SamGuthrie:


SamGuthrie:
On review, this doesn't seem to support live PBE. I had a different PBE version where this worked (preview event) but the paths are different

Morilli:
It's a bit unfortunate that the current code only works for the exporter and not a manual extract of game files. Then again I'm not sure how multiple languages would be supported otherwise.

As for the current pbe path change, I'll make a PR or push to use the new paths after #99 is merged.

PR #102, opened 2024-07-30
Add augments and items to TFT set data

bangingheads:
Adds augments and items per set so that we can distinguish what is current since old TFT data lies around

### DIFF
--- a/cdtb/tftdata.py
+++ b/cdtb/tftdata.py
@@ -182,8 +182,8 @@ class TftTransformer:
                         set_augment_ids.add(item_entries[item_entry].getv("mName"))
                     else:
                         set_items_ids.add(item_entries[item_entry].getv("mName"))
-            set_augments = list(set_augment_ids)
-            set_items = list(set_items_ids)
+            set_augments = sorted(set_augment_ids)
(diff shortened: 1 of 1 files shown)


cdtb/tftdata.py
```diff
@@ -165,7 +169,23 @@ def parse_sets(self, map22, character_names, traits):
...
+                item_list_entry = item_lists[item_list]
+                for item_entry in item_list_entry.getv("mItems", []):
+                    if item_entry not in item_entries:
+                        continue
+                    if item_entries[item_entry].getv("IsAugment"):
+                        set_augment_ids.add(item_entries[item_entry].getv("mName"))
+                    else:
+                        set_items_ids.add(item_entries[item_entry].getv("mName"))
+            set_augments = list(set_augment_ids)
+            set_items = list(set_items_ids)
```
benoitryder, line 186:
```suggestion
            set_augments = sorted(set_augment_ids)
            set_items = sorted(set_items_ids)
```
JSON file will be a bit nicer for humans.

bangingheads: Agreed. Changed.

PR #103, opened 2024-08-04
add strawberry to CDTB

Neinndall:
I added CDTB in the same way as the other modes, strawberry (swarm mode) to get the augments, I'm just starting with python, but it seems to work!
On the other hand, I added strawberry to export as well, just in case it's needed.

### DIFF
--- a/cdtb/export.py
+++ b/cdtb/export.py
@@ -413,6 +413,8 @@ class CdragonRawPatchExporter:
         self.export_tft_data()
         logger.info("export Arena data files")
         self.export_arena_data()
+        logger.info("export StrawBerry data files")
+        self.export_strawberry_data()
 
     def export_tft_data(self):
         if self.patch.version != 'main' and self.patch.version < PatchVersion('9.14'):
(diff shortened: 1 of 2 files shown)


benoitryder:
I know almost nothing about this new game mode, so I won't be able to make relevant comments on the data being generated.
Here are some general remarks (not going into details about the code for now).

* Do we know if Riot plans to make further editions of this mode? If it's only a temporary game mode, it's probably not worth exporting the files. It's nice to have a script as you did but it does not need to be in CDTB. If it's planned to be a major or recurrent mode, why not.
* I think we should use the public name (Swarm) rather than the codename (Strawberry).
* We must not hide errors with `try`/`except`. If something is wrong, we have to handle it correctly. This is important to avoid hiding errors in case format changes (which is very likely to happen if the game mode comes back).
* Looking at the wiki, users will probably expect to have data on game elements other than augments (e.g. boons, items, ...). We don't have to provide everything in a first version but we should provide at least data for all the main elements.

Neinndall:
I had been thinking about it, but after the latest information from Riot, they have a team already dedicated to new game modes, I'm almost sure that this mode will not stay as is, it will be something similar to the Arena mode, that's what I think.

On the other hand, I wanted to make a 1st version getting the augments something similar to the arena mode, although I've been looking at the .bin file carefully and we could add more info!

I could continue reviewing the code and trying to add improvements and try to add more content, obviously we can leave the "pull request" there in case things change, as you see!

Thanks for the answer!

PR #105, opened 2024-11-30
Add unit role to data

stelar7:
Havent tested this yet, but should work in theory

### DIFF
--- a/cdtb/tftdata.py
+++ b/cdtb/tftdata.py
@@ -246,6 +246,7 @@ class TftTransformer:
         champs = {}
         data_characters_dir = os.path.join(self.input_dir, "data", "characters")
         characters_dir = os.path.join(self.input_dir, "characters")
+        role_entries = [x for x in map22.entries if x.type == "TFTCharacterRoleData"]
 
         for champ in champ_entries:
             # always use lowercased name: required for files, and bin data is inconsistent
@@ -261,6 +262,7 @@ class TftTransformer:
 
             tft_bin = BinFile(self_path)
             record = next((x for x in tft_bin.entries if x.type == "TFTCharacterRecord"), {})
+
             if "spellNames" not in record:
                 continue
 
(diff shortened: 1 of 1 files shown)


cdtb/tftdata.py
```diff
@@ -261,6 +262,7 @@ def parse_champs(self, map22, traits):
 
             tft_bin = BinFile(self_path)
             record = next((x for x in tft_bin.entries if x.type == "TFTCharacterRecord"), {})
+
```
benoitryder, line 265: Unneeded change

cdtb/tftdata.py
```diff
@@ -291,6 +293,11 @@ def parse_champs(self, map22, traits):
                 rarity = champ.getv("mRarity", 0) + 1
                 cost = rarity + int(rarity / 6)
 
+            role = "Unknown Role"
+            roleKey = record.getv("CharacterRole", None)
```
benoitryder, line 297:
```suggestion
            role_key = record.getv("CharacterRole", None)
```

cdtb/tftdata.py
```diff
@@ -291,6 +293,11 @@ def parse_champs(self, map22, traits):
                 rarity = champ.getv("mRarity", 0) + 1
                 cost = rarity + int(rarity / 6)
 
+            role = "Unknown Role"
```
benoitryder, line 296: I see no reason for this fallback/default value. We should just use `None`.

benoitryder (changes requested):

Morilli:
I wonder if it could be worth exporting the whole `TFTCharacterRoleData` information as well...

benoitryder:
> I wonder if it could be worth exporting the whole `TFTCharacterRoleData` information as well...

We can start with just the name, then add roles data in a separate top-level entry.

benoitryder:
Replaced by #106

PR #106, opened 2024-12-11
Add unit role to data - with revisions

ColleenN:
Same as https://github.com/CommunityDragon/CDTB/pull/105 but w/ change requests addressed. Thought I'd go ahead and fix it myself since original author hadn't replied in a while and I was the person that asked for the feature in the first place.

### DIFF
--- a/cdtb/tftdata.py
+++ b/cdtb/tftdata.py
@@ -246,6 +246,7 @@ class TftTransformer:
         champs = {}
         data_characters_dir = os.path.join(self.input_dir, "data", "characters")
         characters_dir = os.path.join(self.input_dir, "characters")
+        role_entries = [x for x in map22.entries if x.type == "TFTCharacterRoleData"]
 
         for champ in champ_entries:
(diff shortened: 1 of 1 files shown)


cdtb/tftdata.py
```diff
@@ -291,6 +292,9 @@ def parse_champs(self, map22, traits):
                 rarity = champ.getv("mRarity", 0) + 1
                 cost = rarity + int(rarity / 6)
 
+            role_key = record.getv("CharacterRole", None)
+            role = role_entries[role_key].getv("name", None) if role_key is not None else None
```
benoitryder, line 296:
```suggestion
            role_key = record.getv("CharacterRole")
            role = role_entries[role_key].getv("name") if role_key is not None else None
```
Returning `None` is the default behavior and not explicited in other parts of the code.

benoitryder:
Code was not working due to a small mistake (original code had not been tested).

ColleenN:
Thanks for letting this feature in, it'll be a big help to my data project!!

PR #108, opened 2025-02-01
Lint and cleanup

benoitryder:
Various fixes and improvements I had locally.
I prefer to push in a separate PR, before pushing changes that really matter.

`ruff` use is not enforced but we could decide to add a CI job to help detect dumb mistakes.

### DIFF
--- a/cdtb/storage.py
+++ b/cdtb/storage.py
@@ -21,11 +21,13 @@ class BaseVersion:
     """
 
     def __init__(self, v: Union[str, tuple]):
+        if not v:
+            raise ValueError(v)
         if isinstance(v, str):
             self.s = v
             self.t = tuple(int(x) for x in v.split('.'))
         elif isinstance(v, tuple):
-            self.s = '.'.join(str(x) for x in v)
+            self.s = '.'.join(str(int(x)) for x in v)
             self.t = v
         else:
             raise TypeError(v)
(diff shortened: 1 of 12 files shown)


cdtb/storage.py
```diff
@@ -11,6 +11,7 @@
 
 from .tools import write_file_or_remove
 
+
```
Morilli, line 14: was this intended?

cdtb/storage.py
```diff
@@ -476,3 +479,9 @@ def parse_storage_component(storage: Storage, component: str) -> Union[None, Pat
         return patch
     else:
         return storage.patch_element(name, version, stored=version is not None)
+
+# Import known storages to register them
+# Put the imports at the end to avoid circular dependenceis
+from . import rads as _rads  # noqa
+from . import patcher as _patcher   # noqa
```
Morilli, line 486: What does this do exactly?

Morilli (commented):
Looks good generally, `ruff check` also passes for me locally.

cdtb/storage.py
benoitryder: It imports the storage subclasses so they are registered in `_storage_registry` when the are created (because of the `StorageRegister` metaclass).

cdtb/storage.py
benoitryder: Hum... maybe not. Other files only have 1 empty line before logger import, I'll remove it.

cdtb/storage.py
Morilli: I see, that's a bit of an awkward logic. Looks like `__init__.py` previously did this implicit registering. Fine I guess.

PR #109, opened 2025-02-01
Don't download bundles of already extracted files

benoitryder:
Instead of downloading all bundles then extract everything, download only bundles that are needed to extract the new/changed files.

With this change, bundles are not needed anymore to export older patches (as long as the extracted files there).

Also remove `overwrite` parameter from `PatchStorage.extract_file()`.


Morilli (approved):
Looks sensible.

PR #110, opened 2025-03-21
Include linked files in bin dump

Morilli:
I think it's long overdue for us to include linked file information in bin dumps.

Making this a PR to get feedback on whether the format is acceptable. I don't think this can break compatibility with existing scripts and parsers but another pair of eyes can't hurt.

### DIFF
--- a/cdtb/__main__.py
+++ b/cdtb/__main__.py
@@ -327,6 +327,8 @@ def command_bin_dump(parser, args):
     if args.json:
         json_dump(binfile.to_serializable(), sys.stdout)
     else:
+        if binfile.linked_files is not None:
+            print(f"linked: {binfile.linked_files}")
         for entry in binfile.entries:
             print(entry)
         if binfile.patch_entries is not None:
--- a/cdtb/binfile.py
+++ b/cdtb/binfile.py
@@ -384,6 +384,8 @@ class BinFile:
 
     def to_serializable(self):
         serialized = {entry.path.to_serializable(): entry.to_serializable() for entry in self.entries}
+        if self.linked_files is not None:
+            serialized["__linked"] = self.linked_files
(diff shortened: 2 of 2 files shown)


cdtb/__main__.py
```diff
@@ -327,6 +327,8 @@ def command_bin_dump(parser, args):
     if args.json:
         json_dump(binfile.to_serializable(), sys.stdout)
     else:
+        if binfile.linked_files is not None:
+            print(f"linked: {binfile.linked_files}")
```
benoitryder, line 331: The format should be closer to the syntax used by entries and their fields. And files should be listed one per line (more readable and easier to grep).

Something like this:
```
<Linked [
  file1
  file2
  ...
]>
```
Implementation is easy since we can reuse helpers from `binfile.py`:
```py
slinked = _repr_indent_list(binfile.linked_files)
print(f"<Linked {slinked}>")
```
Moreover, it could be a good idea to move the printing code to a new `BinFile` method.

benoitryder, line 331: Can we find a name a bit more explicit than just "linked"? I don't remember if it links to actual files. If those are used for dependencies, we could use "dependencies".

Morilli: I don't know how exactly these links are used, but `dependencies` probably makes sense. I was mainly following the existing code's naming and what [ritobin](https://github.com/moonshadow565/ritobin/blob/418af7abf5628afb2cfd621f42f70116ef078b0c/README.md?plain=1#L42) uses.

benoitryder: I don't have a strong opinion on this, and you probably know more about those than me. I was just tossing some idea, but we can go with whichever name you choose. ;)

benoitryder (approved):

